PR #4, opened 2025-05-22
Refactor and improve some error handling

pixilcode:
I've created the `feat-reliability` branch in order to track the work that I'm doing. I'm not done yet with everything that I'm doing, but I wanted to give you an opportunity to review and give feedback as I'm working. Each commit should be its own self-contained task for the most part, and some commits have comments giving more details, so I'd recommend reviewing commit by commit, rather than viewing all changes at once. Questions or thoughts?


patrickwalton (commented):
Looks good!

Remind me the reason for using `just` here?

Also, when you raise errors, do those get thrown by the repl or by Python? The goal is for all Oneil errors to get handled by the repl, and in some cases for a debugging state to be entered.

pixilcode:
> Remind me the reason for using `just` here?

`just` is a "recipe runner". Basically, you can put common tasks in the file and run them with `just`. So, for example, I want to run oneil from source, which I do by running `cd src/oneil`, then `python3 -c "import __init__; __init__.main()" <args>`, where `<args>` is the arguments that I want to pass to oneil. This is quite long, and I don't want to have to write it every time that I want to run oneil (since it's pretty frequent). Using `just`, I can instead write `just run <args>` and it runs all of this for me.

Adding this file isn't adding a dependency, this project can be run just fine if we delete it. I was debating whether to put `justfile` into the `.gitignore` file or whether to add it to the project. I decided to add it to the project since, on top of allowing people to run programs, it also adds documentation for how to accomplish useful tasks, such as running oneil from source.

pixilcode:
> Also, when you raise errors, do those get thrown by the repl or by Python? The goal is for all Oneil errors to get handled by the repl, and in some cases for a debugging state to be entered.

That makes sense!

In Python, there are two kinds of exceptions: built-in exceptions and user-defined exceptions.

- **Built-in exceptions** are exceptions that Python uses to indicate that something has gone wrong, such as `ZeroDivisionError` (ex. `5 / 0`) or `ValueError` (ex. `int("abc")`). Usually, these aren't caught inside a program since they indicate a bug in the logic. Instead they raise all the way to the runtime, where they are printed and the runtime exits.
- **User-defined exceptions** are exceptions that are defined in the following way:
  ```py
  class MyError(Exception):
      def __init__(self, my_msg):
          self.my_msg = my_msg
  ```
  These are exceptions that are intended to be raised and caught by the programmer:
  ```py
  try:
      if 1 + 1 != 3:
          raise MyError("1 + 1 is not 3!")
      print("Print me!")                     # doesn't print
  except MyError as err:
      print(f"ERROR: {err.my_msg}")          # prints "1 + 1 is not 3!"
  print("Yay!")                              # prints "Yay!"
  ```

In terms of `raise` and `try: ... except: ...`, here's a good explanation by ChatGPT:

-----

### try: ... except ...

This is used to handle errors (exceptions) that may occur when your code runs.

#### Syntax:

```py
try:
    # Code that might cause an error
except SomeException:
    # Code that runs if that error happens
```

#### Example:

```py
try:
    result = 10 / 0
except ZeroDivisionError:
    print("You can't divide by zero!")
```

In this example, Python tries to run 10 / 0, which would normally crash the program with a ZeroDivisionError. But since it’s inside a try, the program catches the error and runs the except block instead.

### raise

This is used to manually trigger an exception when something goes wrong.

#### Syntax:

```py
raise SomeException(value1, value2, ...)
```

#### Example:

```py
age = -5
if age < 0:
    raise ValueError("Age cannot be negative")
```

Here, you're telling Python to raise an error on purpose if the age is invalid.

### Putting it together

```py
def set_age(age):
    if age < 0:
        raise ValueError("Age cannot be negative")

try:
    set_age(-1)
except ValueError as e:
    print("Caught an error:", e)
```

This prints:

```
Caught an error: Age cannot be negative
```

-----

In terms of `oneil` code, the times when errors are raised, such as [here](https://github.com/careweather/oneil/blob/2e39c4da5dd8b870b641139c0fdb3c348d203d32/src/oneil/__init__.py#L109) and [here](https://github.com/careweather/oneil/blob/2e39c4da5dd8b870b641139c0fdb3c348d203d32/src/oneil/__init__.py#L122), they will be caught by [a handler in the loader](https://github.com/careweather/oneil/blob/2e39c4da5dd8b870b641139c0fdb3c348d203d32/src/oneil/__init__.py#L2693-L2703) where the error is displayed to the user, after which the loader loop continues.

For more details on exceptions, see [Python Docs: Exceptions, sections 8.2-4 and 8.6](https://docs.python.org/3/tutorial/errors.html).

patrickwalton:
Gotcha, so Python won't pick up the Exceptions that are raised, they'll only get picked up if the user defines a try except.

As to just, I think I already have oneil set up to run just by typing 'oneil [args]'. I forget how I implemented that, but did you see that?

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

pixilcode:
> As to just, I think I already have oneil set up to run just by typing 'oneil [args]'. I forget how I implemented that, but did you see that?

Oh, I missed that! I added a section to the docs noting that.

I'll remove the justfile, then.

pixilcode:
> Gotcha, so Python won't pick up the Exceptions that are raised, they'll only get picked up if the user defines a try except.

Not exactly. Whenever an exception is raised, it will propagate up the call stack until it either hits a try except or until it hits the Python runtime.
- if it hits a `try/except`, then it passes the error to the handler. If the error type matches what the handler is catching, then the exception is handled. Otherwise, it keeps going up the call stack.
- if it hits the Python runtime, it is printed out and Python quits. This happens for both built-in and user-defined exceptions.

[This code](https://www.jdoodle.com/ia/1Hwa) has some examples that might be a bit more clear than my explanation.

patrickwalton:
Gotcha.

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

PR #5, opened 2025-05-27
Refactor errors to share a common interface and raise errors rather than returning them

pixilcode:


pixilcode:
> In [this commit](https://github.com/careweather/oneil/pull/5/commits/ed3db7df54fbae29aca8f2af1d1ce33c69bab5f6), it looks like several of the unneeded error handling code lines that you removed had notes. that weren't carried over to the new system. Can you make sure we're not losing any error information that we had?

Yep! I can review all of my changes to see if any information was lost.

pixilcode:
> I noticed unit errors were broken into UnitParseError and UnitError. Maybe we can get more deliberate about the difference? I'm unclear and they seem overlapping to some degree.

In a programming language, there are usually (at least) two different steps, parsing and evaluating.
- `UnitParseError` is an error that occurs at the parsing stage. It says that a given unit is invalid and cannot be parsed.
- `UnitError` is an error that occurs at the evaluation stage. It says that using a valid unit in a given way is an invalid operation.

(In the current implementation, parsing and evaluating are somewhat interleaved. Seperating them makes debugging the language easier and offers more insight into what's going on in each step.)

Maybe it would be helpful to rename `UnitError` to `UnitEvaluationError` to make the difference more clear?

patrickwalton:
Well if we followed Numpy, we'd only return a NaN if dividing zero by zero. Otherwise, dividing a nonzero value by zero would yield inf.

> Finally, I realized that we need to have you testing this on the Veery repository. Have you been doing that? In particular, if you test it on radar.on, you'll be testing most of the other files, since radar.on imports most of them. I see your manual test models, I think we could do a lot more of those tests.

Did you do this? It looks like it erroneously throws an ID_Error on a min() function

patrickwalton:
I should have mentioned! We have two Veery branches you should be testing Oneil against: 0.5.0 and 1.0.0

pixilcode:
> I should have mentioned! We have two Veery branches you should be testing Oneil against: 0.5.0 and 1.0.0

Oh, sweet! Yeah, I just looked at the last commit on the main branch, which is what I've been using, and it turns out it was about 3 months ago :sweat_smile: I am now testing on those branches.

> Well if we followed Numpy, we'd only return a NaN if dividing zero by zero. Otherwise, dividing a nonzero value by zero would yield inf.

Gotcha! I think dividing by zero tends to be a user/logic error that we would want to alert them to. If we return NaN/inf when dividing by 0, in my mind, that would be silently failing, which could produce unexpected results else where that would be hard to trace. (Although if you have experience otherwise, I'm down to discuss further) So I think we should continue to raise an error whenever we hit a division by zero.

> Did you do this? It looks like it erroneously throws an ID_Error on a min() function

I get this error on 0.5.0 `satellite.on`:

```
IDError in model satellite (ID: par_max): Test argument (par_max) uses a reserved keyword.
```

If this is what you're referencing, it turns out that it was an error in how test arguments were parsed. When parsing test arguments, it would filter out the identifiers that it parsed that matched the *original* function (`max`) rather than the *replaced* function (`par_max`), even though the replacement of `max` with `par_max` had already taken place. I've fixed this in the most recent commit.

Aside from that, I ran all the tests against 0.5.0 and 1.0.0, and all of the errors appear to be reasonable errors.

Any other thoughts?

patrickwalton:
> 
> I think dividing by zero tends to be a user/logic error that we would want
> to alert them to.
> 

I think this is fair, though I have seen arguments on the other side. Since this is the state we're at let's leave the error for now.

> 
> If this is what you're referencing, it turns out that it was an error in
> how test arguments were parsed. When parsing test arguments, it would
> filter out the identifiers that it parsed that matched the original function
> ( max ) rather than the replaced function ( par_max ), even though the
> replacement of max with par_max had already taken place.
> 

Interesting, because both 0.5.0 and 1.0.0 work on the 0.13-dev branch. I wonder why the bug only showed up on the new branch.

> 
> Aside from that, I ran all the tests against 0.5.0 and 1.0.0, and all of
> the errors appear to be reasonable errors.
> 

I don't see any errors. Do you? (I see tests that failed, but those aren't errors).

There probably shouldn't be any errors in 0.5.0 unless we change the grammar and need to refactor everything. (Will get back to you on those questions)

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

patrickwalton:
Quick check: did you pull? These branches change frequently and I think the g gnd issue for example was probably already involved

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

pixilcode:
Good question, I just pulled to double-check. `1.0.0` was up to date, `0.5.0` was behind. Running it again on `0.5.0` leaves only the `dynamics_chamber.on` error and the `reflector.on` error. Because `1.0.0` was already up to date, I didn't run it on `1.0.0` again.

patrickwalton:
Okay, gotcha. Well radar (0.5.0) and scatterometer (1.0.0) are the important ones to have working, so I think this PR is good.

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

PR #6, opened 2025-06-03
Add a couple small reliability refactors

pixilcode:
These are a couple more refactorings that I did last week while waiting for feedback. Nothing big, just random things that I noticed.

### DIFF
--- a/src/oneil/__init__.py
+++ b/src/oneil/__init__.py
@@ -7,16 +7,12 @@ import copy
 from beautifultable import BeautifulTable
 import importlib
 from functools import partial
-import traceback, logging
 
 from . import bcolors
 from . import errors as err
 from . import console
 from . import units as un
 from .errors import OneilError
-
-# Configure logging to output to the console
-logging.basicConfig(level=logging.ERROR, format='%(message)s')
 
 np.seterr(all='raise')
 
--- a/src/oneil/__init__.py
+++ b/src/oneil/__init__.py
@@ -56,7 +56,7 @@ def minus(x, y):
         return x - y
     
 
-def parse_file(file_name, parent_model=None):
+def parse_file(file_name):
     parameters = []
     submodels = {}
     imports = []
--- a/src/oneil/__init__.py
+++ b/src/oneil/__init__.py
@@ -350,21 +350,27 @@ def convert_functions(assignment, imports, file_name, line_number):
     pattern = re.compile(r'(\w+)\(([^()]*)\)')
     results = pattern.findall(assignment)
 
-    # If assignment has a function ("name(var1, var2, varn)") in it, replace it with the appropriate callable
-    if results:
-        func, arg_str = results[0]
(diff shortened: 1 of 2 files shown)


patrickwalton:
Looks good, just a couple thoughts:
1. Can you make sure the unused code is unused in the main branch? If not, then it could point to something that got lost in your changes.
2. Can you set up regression testing on radar.on 0.5.0 and scatterometer.on 1.0.0. You probably should pick a specific commit to test against in the process.

One other side thing that I didn't catch from previous changes until later. It looks like the reload command stopped working on your branch. You can test by running a file, leaving the repl open, and changing the file and typing reload in the repl.

pixilcode:
> 1. Can you make sure the unused code is unused in the main branch? If not, then it could point to something that got lost in your changes.

Yep! The `logging` and `traceback` code that I removed was used in `PythonError` to log the Python error, but that is now being handled in a uniform way in the handler/loader. I've attempted to trigger the code in the main branch by raising an error in an imported python function. However, it always gave me a name error and crashed instead.

As for the unused `parent_model` parameter, that was being passed into errors so that the errors could recursively call the loader/interpreter. However, because errors are being raised, this is no longer necessary, so I removed the parameters. I did forget to remove them from other calls to `parse_file` though, so I included a commit to correct that.

> 2. Can you set up regression testing on radar.on 0.5.0 and scatterometer.on 1.0.0. You probably should pick a specific commit to test against in the process.

Done! I've written some scripts that take "snapshots" of the output of oneil and compare against a benchmark snapshot. See [this commit](https://github.com/careweather/oneil/pull/6/commits/5ea3645448aec46795721fb059498a7a187c63da).

> One other side thing that I didn't catch from previous changes until later. It looks like the reload command stopped working on your branch.

Awesome, I fixed this! See [this commit](https://github.com/careweather/oneil/pull/6/commits/bfd528fc9f73dc18402b3c98b8d9d81b27b0ce25).

patrickwalton:
reload fix looks good. Learned how to add comments inline so see those.

pixilcode:
I don't see any comments inline in the code. Did you mean inline somewhere else?

patrickwalton:
Sorry, I should have been more clear. In GitHub's pull request page, you can see comments tied to lines of the diff on the commit. These show up in the discussion tab and the files changed tab.

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

patrickwalton (changes requested):

pixilcode:
Okay, I've updated the regression test so that it *only* uses the output of all model values and all test results. In addition, the snapshots that need to be taken are read from a file.

patrickwalton:
Looks great! Merged it in.

patrickwalton:
Looks great, merged it in. Are we ready to merge the feat-reliability branch you think?

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

PR #10, opened 2025-06-17
Implement the parser for Oneil in Rust

pixilcode:
This is the code for the Oneil parser. I would recommend reviewing the "Files Changed" tab, looking at the final project rather than individual commits. Feel free to annotate with questions about Rust syntax, meaning of code, etc. I've also added instructions to the README so you can experiment a bit locally. I'm also happy at any point to sit down and walk through it with you.

Also, you can mostly ignore the `Cargo.lock` file unless you want to do a full review of all the dependencies in the dependency tree.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "oneil"
-version = "0.1.0"
+version = "0.15.0"
 edition = "2024"
 
 [lib]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,7 +150,7 @@ checksum = "a4895175b425cb1f87721b59f0f286c2092bd4af812243672510e1ac53e2e0ad"
 
 [[package]]
 name = "oneil"
-version = "0.1.0"
+version = "0.15.0"
 dependencies = [
  "clap",
  "nom",


src/rs/ast/mod.rs
patrickwalton, line 1: How does the parameter declaration breakdown described in the readme fit into this? https://github.com/careweather/oneil

Are we redefining any of those? Or have we not got there yet.

src/rs/cli.rs
patrickwalton, line 1: Will the repl not be part of the parser I take it?

src/rs/parser/model.rs
```diff
@@ -0,0 +1,453 @@
...
+            },
+            errors,
+        )))
+    }
+}
+
+/// Attempts to parse declarations
+///
+/// If it fails to parse a declaration, it attempts to recover and continue
+/// parsing. This allows for multiple syntax errors to be found in the model.
```
patrickwalton, line 112: You mentioned descoping this. Note that if still true.

src/rs/parser/token/keyword.rs
patrickwalton, line 1: Where is the error caught if a variable id uses a keyword? Or is that not a problem?

What about for overlapping variable IDs?

Are these part of the parser or come later?

patrickwalton (changes requested):
Just a few comments and questions but mostly makes sense. 

As one added comment, I think a brief developer guide readme would be helpful and it would probably be good to start now and build where you go. The developer guide would just introduce packages, file tree, roles of different files and directories, etc for a new user without all of your programming language intuition.

Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "oneil"
+version = "0.1.0"
```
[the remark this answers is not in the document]
pixilcode: The `edition = 2024` is referencing the edition of the syntax of the Rust programming language, and are using the 2024 edition. See [this page about Rust editions](https://doc.rust-lang.org/edition-guide/editions/index.html) for more details.

Good catch on the version, I forgot to change that from the default. Should I do 1.0.0, or should I do 0.15.0 to give us more flexibility for changes?

src/rs/ast/expression.rs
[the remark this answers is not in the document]
pixilcode: No, `min()` is considered a function call. Binary operators are only infix operators, such as `<expr> + <expr>`, `<expr> - <expr>`, `<expr> * <expr>`, etc.

If you are referring to `BinaryOp::MinMax`, that is the name that I chose for the `|` operator, which produces a minimum and maximum value, such as in the following code example:
```oneil
Upward velocity: v = 5.5 | 17.5 :m/s^2
                     ^^^^^^^^^^
```

src/rs/cli.rs
pixilcode: Yep, that will come later, once we have the module resolver, unit checker, and evaluator.

src/rs/parser/model.rs
pixilcode: Oh yeah, to clarify, I finished this feature, but in the process, I also chose not to continue with other potential features and improvements.

src/rs/parser/token/keyword.rs
pixilcode: > Where is the error caught if a variable id uses a keyword? Or is that not a problem?

In the [parsing function for identifiers](https://github.com/careweather/oneil/pull/10/files#diff-ab548a687f97863565f3e54974a614f8fdfb6e4c6453d29410ed411894158ec3R65-R79), it checks to make sure that an identifier doesn't match any of the keywords.

```rs
pub fn identifier(input: Span) -> Result<Span, TokenError> {
    verify(
        token(
            // ... other code ...
        ),
        // CHECK
        |identifier| !keyword::KEYWORDS.contains(&identifier.fragment()),
    )
    .parse(input)
}
```

> What about for overlapping variable IDs?

This will happen during the module resolution step.

pixilcode:
- [ ] Add a developer guide
- [ ] Add documentation to Expr AST node
- [ ] Reference oneil in cursor rules

Cargo.toml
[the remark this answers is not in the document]
patrickwalton: Sure, yeah, let's do 0.15.0.

src/rs/ast/expression.rs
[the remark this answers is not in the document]
patrickwalton: Gotcha, but minmax was defined as a different function. Is that still included in functions or planned for adding back in?

src/rs/ast/mod.rs
patrickwalton: I was asking how this file fits into the grammar, but I see now that mod.rs is a required rust file.

patrickwalton (approved):

PR #12, opened 2025-07-10
Implement the loader

pixilcode:
I initially planned on squashing these commits into a couple of smaller commits, but it would probably just be easier for you to read through the files changed. If you'd like them squashed, though, I can do that.

Loading models involves:
- parsing each individual model
- loading model dependencies
- converting model parameters and tests to the *intermediate representation* (IR) form
- determining what parameters variables refer to

Note that, in the current implementation, notes and section labels are dropped. This is because the information is not important for evaluation purposes. When we get to adding a VS Code extension and other features, that information will be more relevant, and I'll modify this code to capture it.

Also, all TODO notes in the code are intentionally left incomplete. Some of them are improvements, some of them are bugs, but either way, they are not critical to getting to evaluation level.

### DIFF
--- a/src-rs/oneil_model_loader/src/loader/mod.rs
+++ b/src-rs/oneil_model_loader/src/loader/mod.rs
@@ -224,6 +224,17 @@ fn split_model_ast(
         }
     }
 
+    for section in model_ast.sections {
+        for decl in section.decls {
+            match decl {
+                ast::declaration::Decl::Import(import) => imports.push(import),
+                ast::declaration::Decl::UseModel(use_model) => use_models.push(use_model),
+                ast::declaration::Decl::Parameter(parameter) => parameters.push(parameter),
+                ast::declaration::Decl::Test(test) => tests.push(test),
+            }
+        }
+    }
+
     (imports, use_models, parameters, tests)
 }
 
@@ -292,6 +303,7 @@ mod tests {
     use oneil_ast::{
         Model,
         declaration::{Decl, UseModel},
+        model::Section,
     };
     use std::collections::{HashMap, HashSet};
     use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


src-rs/oneil_model_loader/src/loader/resolver/unit.rs
```diff
@@ -0,0 +1,325 @@
+use oneil_ast as ast;
+
+/// Resolves an AST unit expression into a composite unit representation.
+///
+/// This function takes a parsed unit expression from the AST and converts it
+/// into a `CompositeUnit` that represents the same unit in a flattened form.
+/// The resolution process handles multiplication, division, and exponents
```
patrickwalton, line 7: Where are nonlinear unit operations (dB for example) handled?

test/models/submodel1.on
```diff
@@ -0,0 +1,28 @@
+# Submodel 1: Basic mathematical constants and operations
```
patrickwalton, line 1: Are we at the point where you could also test this on some of the Veery repo files? I'd really like to see you do that since those will be more comprehensive tests.

patrickwalton (approved):
Generally, looks good, but if you could address comments first, that would be great.

docs/decisions/2025-07-02-separate-ast-module-data-structures.md
```diff
@@ -0,0 +1,41 @@
...
+
+This means:
+- IR data structures will have their own `Expr`, `Parameter`, `Limit`, etc. types
+- These types will be defined in the `oneil_ir` crate and its dependencies
+- The IR layer will not depend on `oneil_ast` or `oneil_parser` crates
+- When converting from AST to IR representation, we will transform the data structures
+
+## Consequences
+
+### Positive Consequences
```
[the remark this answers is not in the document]
pixilcode: That makes sense. Maintainability is definitely one of my top goals in designing this implementation. As far as I'm aware, the negative consequences listed are not going to impact maintainability that much for a couple reasons:
1. **The two data structures are slightly different**, even if they look similar on the surface. Specifically, the way that variables are stored differs, since the intermediate representation resolves a variable to its exact location, either in this model or elsewhere. Trying to reuse the AST to represent this would introduce complexity into the code.
2. **The Rust compiler is very good at ensuring that types match up correctly.** If you add a type to the AST, you will be required to handle it in the transformation step from AST to IR. If the translation between the two is direct (such as "AST binary op" -> "IR binary op"), this is simple and only requires a couple extra lines of code.

The benefits greatly outweighed the negatives and I didn't see this decision as a big decision, which is why I didn't check with you. My main reason for documenting it was that it might be confusing for someone new looking at the code. This document was mostly written by AI, and I just looked it over to make sure it sounded right.

If you would like, I can edit it and add more depth and detail. My main goal in writing it was just to have a quick piece of documentation that approximated my thoughts. In addition, I will make sure that if I need to make maintainability trade-offs, I'll talk to you about them.

src-rs/oneil_model_loader/src/loader/resolver/unit.rs
pixilcode: All units will be converted into useable units in the next step (unit checking). This includes checking for prefixes such as `dB` or `k`, as well as checking for custom units such as `N` or `W`.

test/models/submodel1.on
pixilcode: Yeah, I can test on Veery repo files. I'll make a branch off of the one used for regression testing and update the notes and strings and report back on the results. There may be features missing that will cause it to fail, and I'll report on that as well.

docs/decisions/2025-07-02-separate-ast-module-data-structures.md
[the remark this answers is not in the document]
patrickwalton: No need to do anything else. I appreciated the thought process and think your decision makes sense.

PR #13, opened 2025-07-29
Improved error messages

pixilcode:
Error messages have been updated to be readable. They generally follow the following style

```
error: unclosed `(`
 ---> foo.on:15:13
   |
15 | X: x = (1 + 3
   |              ^
```

This also required changes in the structure of parser errors, IR errors, and the IR itself.


pixilcode:
Okay, I believe it is now ready for review.

patrickwalton (changes requested):
Since you did all that testing on the actual files, I'm going to take that as review. Go ahead and make the changes to the model we discussed over email and then test everything again. Once there are no error messages on your veery model branch, we can go ahead and call this PR done.

pixilcode:
Alright, I've implemented parser changes that we discussed in the parser email:
- tests are allowed to have notes
- `1`, `%`, and `$` are accepted as units
- duplicate imports, submodel uses, and parameters are disallowed
- numbers starting with a decimal point (ex. `.123`) are allowed
- label character support was extended to accept more characters
- chained comparison operators are allowed

In addition, I removed support for test inputs as you mentioned in office a couple weeks ago (ex. `test {a, b}: a < b` and `use foo(a = 1, b = 2)`).

Now that all of the parsing errors have been resolved, I'm running into some resolution errors. I've emailed you with the specific problems, but there are module resolution and duplicate parameter issues.

PR #18, opened 2026-01-12
Goto definition implementation

pgattic:
Basic working implementation of Goto definition. Doesn't yet work across models.

### DIFF
--- a/src-rs/oneil_lsp/src/lib.rs
+++ b/src-rs/oneil_lsp/src/lib.rs
@@ -130,8 +130,7 @@ impl LanguageServer for Backend {
             .await;
 
         if let Err(error) = result {
-            let _ = self
-                .client
+            self.client
                 .log_message(MessageType::ERROR, format!("did_change error: {error}"))
                 .await;
         }
@@ -186,20 +185,21 @@ impl LanguageServer for Backend {
         let uri = params.text_document_position_params.text_document.uri;
 
         // Convert LSP position to byte offset
-        let offset = match self.docs.position_to_offset(&uri, position).await {
-            Some(offset) => offset,
-            None => {
-                self.client
-                    .log_message(MessageType::WARNING, "Could not convert position to offset")
-                    .await;
-                return Ok(None);
-            }
+        let Some(offset) = self.docs.position_to_offset(&uri, position).await else {
+            self.client
+                .log_message(MessageType::WARNING, "Could not convert position to offset")
+                .await;
+
+            return Ok(None);
         };
 
         self.client
             .log_message(
                 MessageType::INFO,
-                format!("goto_definition: offset={}, position={}:{}", offset, position.line, position.character),
+                format!(
+                    "goto_definition: offset={}, position={}:{}",
+                    offset, position.line, position.character
+                ),
             )
             .await;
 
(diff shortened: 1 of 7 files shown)


src-rs/oneil_lsp/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+
+#[tokio::main]
+pub async fn run() {
+    let stdin = tokio::io::stdin();
+    let stdout = tokio::io::stdout();
+
+    let docs = Arc::new(DocumentStore::new());
+    let (service, socket) = LspService::new(|client| Backend {
+        client,
+        docs: docs.clone(),
```
pixilcode: nit: if you're intending to clone the `Arc` (which I assume is your goal), it's better to state that explicitly with `Arc::clone`, since that makes it clear that you're cloning a reference, not the underlying data.

src-rs/oneil_lsp/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+        let result = self
+            .docs
+            .apply_changes(params.text_document, params.content_changes)
+            .await;
+
+        if let Err(error) = result {
+            let _ = self
+                .client
+                .log_message(MessageType::ERROR, format!("did_change error: {error}"))
+                .await;
```
pixilcode: nit: you can probably get rid of the `let _ =` since this just returns `()` anyway

src-rs/oneil_lsp/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            &file_parser::FileLoader,
+        ) {
+            Ok(collection) => collection,
+            Err(_) => {
+                self.client
+                    .log_message(MessageType::ERROR, "Failed to load model")
+                    .await;
+                return Ok(None);
+            }
+        };
```
pixilcode: clippy nit: this could be rewritten using `let Ok(model_collection) = oneil_model_resolver::load_model(...) else { ... }` to make the happy path more clear

src-rs/oneil_lsp/src/symbol_lookup.rs
```diff
@@ -0,0 +1,281 @@
...
+
+            // Check references
+            if let Some(reference) = current_model.get_references().get(&ir::ReferenceName::new(model_name.clone())) {
+                let external_model = model_collection.get_models().get(reference.path())?;
+                let param = external_model.get_parameter(parameter_name)?;
+                return Some(span_to_location(reference.path(), param.name_span()));
+            }
+
+            None
+        }
```
pixilcode: For this to work the way it is expected to, I believe you'll want to match `ExternalReference { model_path, .. }`, then just return something like `Some(span_to_location(model_path, None))`, where `None` indicates the start of the file. It could also be like the match branch for `ModelImport { path, .. }`.

As a side note, I believe it would make more sense for this to go to the `ref` or `use` declaration in the same file. The user can then use that to jump to the other model if that's what they desire. (That's how other languages implement this feature). This might be a task for later, though.

pixilcode (changes requested):
Overall, it looks good! There are a couple clippy lints that you could change, as well as one potential bug in the implementation. Any comments marked with `nit` aren't  crucial to change, just code style.

I'm not sure if we want to address the bug in this pull request, though. Jump to definition works great as long as you aren't looking for the definition of a model reference. And in that case, it just can't find the definition. Fixing this could be a task for later.

I know you're pretty busy with school, so let me know if you need me to do wrap any/all of these up.

PR #26, opened 2026-03-05
feat: Add ability to import Python functions

pixilcode:
This pull request makes Oneil capable of importing Python functions.

It also fixes some differences between the old Oneil implementation output and the current Oneil implementation output by comparing outputs based on the radar model in `veery/model/radar.on` (v0.5.0). After modifying the model to fix breaking changes, the following outputs still differ. (`-` indicates that it was present in the old Oneil, `+` indicates that it is present in the new Oneil)

## Parameters (v0.5.0)

```py
# This differs because of improved `sin` calculation with intervals

- v_r = -0.0 :m/s  # Radial look speed
+ v_r = -5017.757065694405 | 5017.757065694405 :m/s  # Radial look speed


# These differ because of other improved interval operations (as far as I can tell)

- B_Dc = 50.04 :kHz  # Doppler bandwidth across a single measurement cell
+ B_Dc = 50.04990304637924 | 50.04990304637924 :kHz  # Doppler bandwidth across a single measurement cell

- U_P = 13.99 | 62.4 :%  # Orbital average power usage
+ U_P = 13.990876122033699 | 63.52735389931959 :%  # Orbital average power usage

- U_P1 = 115.1 | 448.1 :%  # Always-on orbital average power usage
+ U_P1 = 115.12733291267396 | 456.1863041344151 :%  # Always-on orbital average power usage
```

## Tests (v0.5.0)

```py
# These differ because of improved interval operations

  test (satellite): max(f_up.t + f_DTru) < max(f_upa.t)
    result: fail
-   - f_DTru = -3.508 | 3.519 :kHz
+   - f_DTru = -3.5194366845281886 | 3.5194366845281886 :kHz
    - f_up = 148.9 | 149.4 :MHz
    - f_upa = 149.1 | 149.2 :MHz

  test (satellite): min(f_up.t + f_DTru) > min(f_upa.t)
    result: fail
-   - f_DTru = -3.508 | 3.519 :kHz
+   - f_DTru = -3.5194366845281886 | 3.5194366845281886 :kHz
    - f_up = 148.9 | 149.4 :MHz
    - f_upa = 149.1 | 149.2 :MHz

# This one differs because in the past, Oneil did
#   `a_min | a_max > b_min | b_max`
# as
#   `a_min > b_min && a_max > b_max`
# but now it does
#   `a_min > b_max`
# since that ensures that `a` will always be greater than `b`
  test (nest): t_op > t_b
-   result: pass
+   result: fail
+   - t_b = 1.0 | 6.0 :hr
+   - t_op = 5.032652327601702 | 8.579307527321008 :hr

# I updated this one because the author intention seemed
# to be to compare the mins and the maxes. I can always
# change this one back, though. It will just fail for the same
# reason as the previous one
- test (radar): r_g <= a_g
+ test (radar): min(r_g) <= min(a_g) and max(r_g) <= max(a_g)
    result: pass

# This one passed with the old Oneil because string comparison
# didn't work. Even though `O_r` is `footprint`,
# `O_r != 'footprint'` because in the `__ne__` implementation
# for `Parameter`, comparing to a string returns `self.min == other`
# instead of `self.max != other`. Regardless, this is fixed in the new
# Oneil, which causes this test to fail
  test (radar): r_r >= r_rrd or O_r != 'footprint'
-   result: pass
+   result: fail
+   - O_r = 'footprint'
+   - r_r = 38.38727203379075 | 40.98879236334264 :km
+   - r_rrd = 3006786.5049919235 | 3060933.7968768864 :m
```

I haven't yet worked through `v1.0.0` differences, but I imagine that I have found most of the implementation differences working through `v0.5.0`. In my opinion, it would probably be more worth it to move on to the next task since working through `v1.0.0` will probably take another day or two at least, and I'll have to work through the differences again when we move everyone over to Oneil 0.15.0. However, if you'd like, I can do `v1.0.0` as well.


pixilcode:
We decided that the differences would be reviewed again when I actually convert the veery 0.5.0 model, so I will merge this as complete.

PR #28, opened 2026-03-08
[Refactor] Move shared types to oneil_shared

TimWhiting:
These 6 types were defined identically in both oneil_ast and oneil_ir. Now they live in oneil_shared with re-exports from both crates. The resolver's identity-mapping functions are simplified to simple derefs. Trivial constructors are also inlined into the call sites.


pixilcode:
Oh, sorry, I didn't mean to close these, I was just cleaning up branches that I thought were unused. Some of these (`Identifier` and `Label`) were done in a previous commit, but it could be helpful to move the others into the shared crate as well. If you'd like, you're welcome to update this and submit a pull request to proj/rust-rewrite! No worries if not though.

PR #31, opened 2026-03-19
feat: Completed CLI

pixilcode:
This is the working CLI for Oneil.

To install follow the instructions in `docs/guide/01-installation.md`. For commands that can be run, see `oneil --help`. The main one that you'll be using is `oneil eval <model-file>`. Let me know if you have any questions!


patrickwalton:
Working on testing/reviewing. As a side note, can you take a note in airtable or issues that we will want to update VIM highlighter with the new grammer at some point.

patrickwalton:
I tested this on the branch in https://github.com/careweather/veery/pull/449

The first thing I notice is that you haven't set a default number of significant figures. We need to figure out how to handle that. It looks like the original oneil defaults to 4, but I believe that can be configured as an argument. This makes the output practically unreadable if not handled correctly. Update your original comment with the new outputs once you've addressed this.

Another thought is that I like that Python CLI doesn't require you to use an argument to run a file. we should have a default assumption about what `oneil <file-name>` does. In the current oneil, it runs the repl, but I assume in the new oneil it could be the equivalent of `oneil eval <filename>`.

In general, I'm not too worried about the values that are 5-10% off. Your calculation is probably better, so I'll chalk the changes up to that, but having this record of the regression test is good if we have questions in the future.

I like the new tree, but it's hard to use with no levels cutoff option.

The `independent` command doesn't seem to work.

```
[patrick@anubis ~]$ oneil independent radar
[patrick@anubis ~]$ oneil independent solar.on
[patrick@anubis ~]$ oneil independent
error: the following required arguments were not provided:
  <FILE>

Usage: oneil independent <FILE>

For more information, try '--help'.
[patrick@anubis ~]$ 
```

> These differences are caused by the duplicate names solution that you proposed:
> https://github.com/careweather/veery/commit/e2cb01104a29bf9176b3efd9b2c835244a8b1789

Can you clarify?

> Added to resolve namespace problem
+ S_tauIQ = 65.536 :kB  # Pulse size

Clarify this as well?

> This test appeared to be a component-wise comparison rather than an interval comparison,
> so I updated the test

> - test (radar): r_g <= a_g
> + test (radar): min(r_g) <= min(a_g) and max(r_g) <= max(a_g)
>     result: pass

I'll need to think about this. I'll also need to think a bit more about v_r. While the new calculation is probably correct, it probably means we need to reimplement the model to get the old result. I'll come back to this once you've addressed the other comments above.

pixilcode:
> The first thing I notice is that you haven't set a default number of significant figures. We need to figure out how to handle that. It looks like the original oneil defaults to 4, but I believe that can be configured as an argument. This makes the output practically unreadable if not handled correctly. Update your original comment with the new outputs once you've addressed this.

Sounds good, I don't think that will be too hard to update.

> Another thought is that I like that Python CLI doesn't require you to use an argument to run a file. we should have a default assumption about what `oneil <file-name>` does. In the current oneil, it runs the repl, but I assume in the new oneil it could be the equivalent of `oneil eval <filename>`.

Okay, I'll look into that and see how to get that to work.

> I like the new tree, but it's hard to use with no levels cutoff option.

There is `oneil tree radar.on a_r --depth 3`, which only shows up to three levels of dependencies. Right now, `tree` shows all dependencies within the model, but I could make it so that it defaults to a given depth if that would be preferable.

> The `independent` command doesn't seem to work.

Odd... I'll look in to that.

> > These differences are caused by the duplicate names solution that you proposed:
> >     https://github.com/careweather/veery/commit/e2cb01104a29bf9176b3efd9b2c835244a8b1789
> 
> Can you clarify?
> 
> >     Added to resolve namespace problem
> > 
> >     S_tauIQ = 65.536 :kB # Pulse size
> 
> Clarify this as well?

When I was doing my original refactoring of Veery 0.5.0 (the old version that we ended up discarding), I ran into namespace conflicts, which you fixed in the given commit: https://github.com/careweather/veery/commit/e2cb01104a29bf9176b3efd9b2c835244a8b1789. I applied these changes to the new refactorings to fix the namespace problems, and they changed the output of the new Oneil.

> >     This test appeared to be a component-wise comparison rather than an interval comparison,
> >     so I updated the test
> > 
> >         test (radar): r_g <= a_g
> > 
> >         test (radar): min(r_g) <= min(a_g) and max(r_g) <= max(a_g)
> >         result: pass
> 
> I'll need to think about this. I'll also need to think a bit more about v_r. While the new calculation is probably correct, it probably means we need to reimplement the model to get the old result. I'll come back to this once you've addressed the other comments above.

Sounds good! Let me know what you decide on this. I think I should be able to reproduce the old results pretty easily.

pixilcode:
Okay, I've updated it so that sig figs are customizable with a `--sig-fig` flag, but sig figs default to 4. This is specifically sig figs _after_ the decimal point. If you were looking for something else, let me know and I can figure it out.

I've also figured out how to get it to default to the `eval` command when no subcommand is provided.

pixilcode:
Doing a bit of further research into how Python prints values, I could try implementing [the "Dragon2" or "Dragon4" algorithms that Python uses](https://michael-koller-91.github.io/blog/2023-03-26-how-to-print-floating-point-numbers.html) if you'd like. That would provide more accurate outputs, and I believe that it would allow me to actually just get 4 sig figs. I could also try and implement [one of the other floating point printing algorithms](https://github.com/abolz/Drachennest).

patrickwalton:
> but I could make it so that it defaults to a given depth if that would be preferable.

Let's default to 4 as in the original Oneil. I'd like us to still show levels are clipped as the original Oneil does (with turtles), but I realize that's not self-explanatory, so if you can think of a more standard notation to show recursion is clipped, please propose it.

> This is specifically sig figs after the decimal point. If you were looking for something else, let me know and I can figure it out.

How does the original Oneil implement it? You haven't clarified why you're reinventing the wheel here. Significant figures isn't usually decimal point dependent. Look at all of the original results in https://github.com/careweather/veery/pull/449. Notice how there are always four digits given regardless of the size of the value

0.008534 (this should really default to 8.543e-3)
0.1707
4.928
15.21
511.3
5707.0
2.614e5
5.216e12

If those other algorithms are easier and more consistent for implementing what we had before, then go for it.

patrickwalton:
I tried to dig into the v_r difference, but I don't have a way to test a calculation like I would in the REPL. Do you think it would make sense to have an expression eval option in the CLI? Or should we just wait for the REPL? If we want to wait, I can just test by writing files and eval'ing them.

> test (radar): r_g <= a_g
> test (radar): min(r_g) <= min(a_g) and max(r_g) <= max(a_g)
result: pass

Your fix is correct.

patrickwalton:
Okay, I went ahead and tested by just writing a test file that evaluates partial expressions and I understand better now. I think the v_r difference is a major bug that your improved implementation solves. Good work! However, I need to evaluate the downstream effects and I don't see the dependents command in the --help for oneil cli. Can you add it?

pixilcode:
> Let's default to 4 as in the original Oneil. I'd like us to still show levels are clipped as the original Oneil does (with turtles), but I realize that's not self-explanatory, so if you can think of a more standard notation to show recursion is clipped, please propose it.

Sounds good. I'll ponder more on how to display that.

> How does the original Oneil implement it? You haven't clarified why you're reinventing the wheel here.

Yeah, I left out a bit of context, sorry. I realized after I'd implemented it that you _were_ in fact looking for something else. After doing research, I found that the original Oneil uses Python's builtin float formatter, which is implemented using the algorithm mentioned above (Dragon 2). So if we wanted to replicate original Oneil output, we'd need to implement that same algorithm. That algorithm has not yet been implemented anywhere in Rust, at least not as far as I could find, so I'd need to reimplement it from scratch.

For reference, this is essentially where the numbers are formatted.

```py
def _round(num, n=3):
    formatstr = "%." + str(n) + "g"
    return float(formatstr % num)
```

Note the format string, `%.3g`. Rust doesn't have an equivalent formatting flag, but I may be able to approximate it using standard float formatting and another `e` flag that I found that I think has similar behavior for scientific notation. I'll see what I can do.

> I tried to dig into the v_r difference, but I don't have a way to test a calculation like I would in the REPL. Do you think it would make sense to have an expression eval option in the CLI? Or should we just wait for the REPL? If we want to wait, I can just test by writing files and eval'ing them.

I ran into this same problem when I was working on migrating `veery`, so I added `--expr`/`-x` option to `oneil eval`. So you can do something like `oneil eval model.on -x 'a + b'` and it will print the result of evaluating the expression in the context of the given model.

> Okay, I went ahead and tested by just writing a test file that evaluates partial expressions and I understand better now. I think the v_r difference is a major bug that your improved implementation solves. Good work! However, I need to evaluate the downstream effects and I don't see the dependents command in the --help for oneil cli. Can you add it?

Yay! :tada: Based on our conversation (a while back) about how `dependents` and `tree` are similar in concept, I implemented the `dependents` command using the `--list-refs` option for `oneil tree` (so `oneil tree model.on param --list-refs`), but I can easily make it a seperate command if you'd like.

pixilcode:
(Also, I chose `--list-refs` since it can be confusing to talk about "dependencies" and "dependents", so I tried to make the difference a little more clear by using "references" instead of "dependents". We don't have to rename the `dependents` command to `references`, I'm just explaining the rational behind the name of the flag)

patrickwalton:
> I ran into this same problem when I was working on migrating veery, so I added --expr/-x option to oneil eval. So you can do something like oneil eval model.on -x 'a + b' and it will print the result of evaluating the expression in the context of the given model.

Great! Make sure to add to the --help! Or to the man page when we package.

> Also, I chose --list-refs since it can be confusing to talk about "dependencies" and "dependents"

Hmm...I saw --list-refs and glossed over it, so I'm not sure the name is clearer. But I think maybe the fix is to combine it into tree and have the user specify direction (up/down). What do you think? We wouldn't need to consolidate them right away. Could push later.

pixilcode:
> Hmm...I saw `--list-refs` and glossed over it, so I'm not sure the name is clearer. But I think maybe the fix is to combine it into tree and have the user specify direction (up/down). What do you think? We wouldn't need to consolidate them right away. Could push later.

Yeah, in retrospect `--list-refs` does seem pretty easy to miss since it isn't exactly clear what `refs` means. I was trying to make it short while also making it clear what the flag does, but I don't think I accomplished either...

Using directions could work. We could make it into a flag (`--up`/`--down`) or we could make it into a subcommand (`oneil tree up`/`oneil tree down`).

We could also make the `--list-refs` flag into a long/short combo (ex. `--list-references`/`-R`) so that it's clear what it means, but there's also an option for typing it out quickly.

We could also just make two different commands (ex. `oneil dep-tree`/`oneil ref-tree`) which would make it clear that there are two different kinds of trees rather than hiding that fact behind an optional option that's easy to miss.

I'm down to implement whichever option you'd prefer. And if you'd like to shift it to a later task, I can do that too.

patrickwalton:
> Using directions could work. We could make it into a flag (--up/--down)

Do this, but make it default to up if no option is given.

pixilcode:
Just to be clear, what is your definition of up? Is it finding dependencies or is it finding references?

patrickwalton:
Good question. Up would show dependents. Down would show parameters that are used to calculate the given equation (current tree).

So actually down should be the default.

pixilcode:
Okay, I've added sig fig handling that is similar to `%g` in Python and I replaced `--list-refs` with `--up` and `--down`.

patrickwalton:
Let me know when you feel all comments on both PRs is addressed.

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

pixilcode:
> The independent command doesn't seem to work.

Did you ever get this working? It Works On My Machine™, so if it isn't working on yours, I'll have to figure out how to debug it.

> I'd like us to still show levels are clipped as the original Oneil does (with turtles), but I realize that's not self-explanatory, so if you can think of a more standard notation to show recursion is clipped, please propose it.

Okay, I've come up with a proposition for this. What do you think of either

```
├── theta_32.a = 0.8536 | 1.028 : rad
│   │          = theta_3(lambd, L_2, alpha_2)
│   └──╶╶╶
```
or
```
├── theta_32.a = 0.8536 | 1.028 : rad
│   │          = theta_3(lambd, L_2, alpha_2)
│   └──╶╶╶ (clipped)
```
where `(clipped)` is grayed out?

patrickwalton:
> The independent command doesn't seem to work.

I got it to work. I was in the wrong directory before :facepalm: 

I really like your first proposition for clipping tree. I think it's self-explanatory and concise.

pixilcode:
Sweet! I believe this PR is good to go.

patrickwalton:
Do you have instructions for how to update installs? I assume it's dependent on install method?

pixilcode:
I just added them to the user guide. Basically, just update the source code with `git`, then re-install Oneil.

patrickwalton:
Can you add shorthands for the up and down cli arguments? 

```
[patrick@anubis model]$ oneil tree -u radar
error: unexpected argument '-u' found

  tip: to pass '-u' as a value, use '-- -u'

Usage: oneil tree [OPTIONS] <FILE> <PARAM>...

For more information, try '--help'.
```

Also, I'm not quite sure what the tip is implying?

patrickwalton:
I regret my thinking about the default for recursion. I think we should let the user choose to clip the output. My bad.

Can we flip the up tree? Have the requested value at the bottom and have it branch upwards (like a family tree on familysearch)

```
[patrick@anubis model]$ oneil tree --up radar v_r
radar.on v_r = -7622 | 7622 : m/s
│            = v.o*cos(phi_e2)*cos(psi)
├── radar.on f_D = -21.28 | 21.28 : kHz
│                = v_r/lambd.a
└── radar.on f_De = -42.56 | 42.56 : kHz
    │             = 2*v_r/lambd.a
    ├── radar.on B_D = 85.11 : kHz
    │   │            = max(f_De)-min(f_De)
    │   ├── radar.on B_Dc = 85.11 | 85.11 : kHz
    │   │   │             = B_D*r_xx//a_r
    │   │   └── radar.on B = 85.16 : kHz
    │   │       │          = B_t + 2*max(abs(f_De))
    │   │       └──╶╶╶
    │   └── radar.on B = 85.16 : kHz
    │       │          = B_t + 2*max(abs(f_De))
    │       ├── radar.on n_bp = 7.515 : MHz/kHz
    │       │                 = S_f*(f_adc/B)
    │       ├── radar.on L_D = 1
    │       │   │            = max(1, B/B_r)
    │       │   └──╶╶╶
    │       └── radar.on B_x = 22.26 : kHz
    │                        = B_r - B
    └── radar.on B = 85.16 : kHz
        │          = B_t + 2*max(abs(f_De))
        ├── radar.on n_bp = 7.515 : MHz/kHz
        │                 = S_f*(f_adc/B)
        ├── radar.on L_D = 1
        │   │            = max(1, B/B_r)
        │   └── radar.on L_Rx = 4.086 | 5.414 : dB
        │       │             = L_r*L_sp*L_tau*L_D
        │       └──╶╶╶
        └── radar.on B_x = 22.26 : kHz
                         = B_r - B
```

pixilcode:
> Can you add shorthands for all of our cli arguments?

Yep!

> Also, I'm not quite sure what the tip is implying?

It means that if you want to pass `-u` as a value, not as a flag, then you need to include a `--` before it. So, if for some reason you have a model named `-u.on`, you could eval it with `oneil eval -- -u`. In other words, all values after a `--` are treated as a value rather than a flag or an option.

It's a tip that comes by default with the CLI argument parsing library that Oneil uses. I don't think there's a way to disable it.

> I regret my thinking about the default for recursion. I think we should let the user choose to clip the output. My bad.

Sounds good, that's not to hard to change!

> Can we flip the up tree? Have the requested value at the bottom and have it branch upwards (like a family tree on familysearch)

Yep, I should be able to do that.

Also, side note, did you mean to close the PR?

patrickwalton:
Got it. CLI library tip makes a ton of sense now.

No I must have clicked close on accident. I see how to reopen now.

pixilcode:
Okay, I've finished working on the tree work. It's back to having no default depth, and trees are now printed upside down:

```
        │   ┌── phi = 0 | 10 : deg
        │   │       = max(0, min(phi_t) - max(e_phi))|max(phi_t) + max(e_phi)
        │   │   ┌──╶╶╶
        │   ├── theta_31.a = 0.9756 | 1.089 : rad
        │   │              = theta_3(lambd, L_1, alpha_1)
        ├── phi_e2 = -31.2 | 41.2 : deg
        │          = min(phi - 0.5*theta_31.a) | max(phi + 0.5*theta_31.a)
        ├── R_E.c = 6.357e6 | 6.378e6 : m
        │         = 6356752.314|6378137.0
        ├── h.o = 510 : km
        │       = 510
    ┌── R_sr = 510 | 733 : km
    │        = h.o|sin(max(abs(alpha_e)))*R_E.c//sin(max(abs(phi_e2)))
┌── R_s = 510 | 733 : km
│       = R_sr * 1
│   ┌──╶╶╶
├── theta_32.a = 0.8536 | 1.028 : rad
│              = theta_3(lambd, L_2, alpha_2)
├── R_E.c = 6.357e6 | 6.378e6 : m
│         = 6356752.314|6378137.0
a_g = 466.3 | 847.2 : km
    = a_g(theta_32.a, R_s, R_E.c)
```

For the shortened values, here's what I've got.

| command | shortened |
|-----------------|----------------|
| --sig-figs | -s? -f? |
| --no-colors | none, it probably won't be used frequently, mainly just for CI
| --venv-path | none, this is also unlikely to be used, it's mainly for configuring the LSP |
| --params | -p |
| --print | -P |
| --debug | -D |
| --watch | -w |
| --expr | -x |
| --recursive | -r |
| --partial | ? |
| --no-header | none, this is likely only going to be used for regression testing? |
| --no-test-report | none, see --no-header |
| --no-parameters | none, see --no-header |
| --down | -d |
| --up | -u |
| --depth | ? |
| --values | none? -v is typically used for verbose tracing and -V is typically --version. Also, this is only in the `independent` command for showing the values of independent variables |

I think that's all of them. Thoughts?

patrickwalton:
Looks great!

I edited my shortened request to just up/down. sorry I should have messaged you.

Here are my thoughts:

| command | shortened |
|-----------------|----------------|
| --sig-figs | none, easier to leave the ones that take values (like the number of sig figs) with no short form |
| --no-colors | none, it probably won't be used frequently, mainly just for CI
| --venv-path | none, this is also unlikely to be used, it's mainly for configuring the LSP |
| --params | -p |
| --print | -P |
| --debug | -D |
| --watch | -w |
| --expr | -x |
| --recursive | -r |
| --partial | <what does this do?> |
| --no-header | none |
| --no-test-report | none, see --no-header |
| --no-parameters | none, see --no-header |
| --down | -d |
| --up | -u |
| --depth | none, easier to leave the ones with arguments with no short |
| --values | let's just default to showing values for independent and remove this option |

pixilcode:
`--partial` enables partial evaluation. In other words, it allows you to print out results even if there are errors. For example, given

```
# foo.on
X: x = 1 :m
Y: y = 1 :s 
Z: z = x + y :m
```

`oneil foo.on --params x` outputs

```
error: expected unit `m` but found `s`
 --> /tmp/on.on:3:12
  | 
3 | Z: z = x + y :m
  |            ^


note: expected because this expression has unit `m`
 --> /tmp/on.on:3:8
  | 
3 | Z: z = x + y :m
  |        ^
```
while `oneil foo.on --params x --partial` outputs

```
error: expected unit `m` but found `s`
 --> /tmp/on.on:3:12
  | 
3 | Z: z = x + y :m
  |            ^


note: expected because this expression has unit `m`
 --> /tmp/on.on:3:8
  | 
3 | Z: z = x + y :m
  |        ^


────────────────────────────────────────────────────────────────────────────────
Model: /tmp/on.on
Tests: 0/0 (PASS)
────────────────────────────────────────────────────────────────────────────────
x: x = 1 : m  # X
```

because `x` was calculated without any errors.

`--partial` was just the name that I came up with because of "partial evaluation". I realize that may be unclear to the average user, though. I'm happy to rename it if you have other ideas for a name.

patrickwalton:
How is partial different from debug?

pixilcode:
Great question! `--debug` has to do with debug annotations (`**`).

In the Python version, debug (`**`) and trace (`*`) annotations start a debugger at certain places in the code. I couldn't start a debugger in the middle of Rust code, and I didn't think it would be helpful in the moment to work on a debugger, so I instead decided to approximate similar helpful features.

So `*` marks a parameter as a trace parameter, which means that it gets printed out when printing `--print trace`. `**` marks a parameter as a debug parameter, which also gets printed out with `--print trace`. In addition, whenever `--debug` is used, the direct dependencies of the debug parameter are printed (similar to `oneil tree`, though this was before `oneil tree` existed).

```oneil
────────────────────────────────────────────────────────────────────────────────
Model: model.on
Tests: 0/0 (PASS)
────────────────────────────────────────────────────────────────────────────────
z = 2 : m  # Z
  - x = 1 : m
  - y = 1 : m
```

So that's what the `--debug` flag does.

Now that I've answered your question, I think that as they currently exist, we could probably get rid of trace and debug annotations (unless you have other ideas for how they can be used). I can easily print out whatever parameters I'd like without marking them as trace by using `oneil model.on --params x,y`, and I can print out dependencies without marking them as debug by using `oneil tree model.on z`. That's what I ended up doing while I was figuring out Veery 0.5. (Also, they aren't currently used in `veery` anywhere)

If we get rid of trace annotations, we could get rid of `--debug` (or replace `--partial` with `--debug` if you think that works better). We could also probably get rid of `--print <mode>` and just have a `--print-all-params` or something like that, since I only see that option being used in regression testing. That would simplify some of the code as well.

What do you think?

patrickwalton:
I was just thinking we could combine --partial and --debug functionality under --debug for now.

pixilcode:
Done! Is there anything else you'd like me to do?

PR #33, opened 2026-04-01
Install updates

pixilcode:
This PR
- improves the error messages in the install script when `gcc` or `python-devel` is not installed
- adds a minimum Rust version so that compiling with lower Rust versions doesn't give confusing errors
- updates crate versions to the newest ones available to improve performance

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -908,9 +908,9 @@ dependencies = [
 
 [[package]]
 name = "pyo3"
-version = "0.28.0"
+version = "0.28.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


patrickwalton:
Anything specific you wanted me to review? If not, you can go ahead and merge.

pixilcode:
I looked into [the `tokio` score for "Supply Chain Security"](https://socket.dev/cargo/package/tokio), and it looks like it is because it has code that accesses the network, which is part of what `tokio` is supposed to do. I think it should be okay.

pixilcode:
> Anything specific you wanted me to review? If not, you can go ahead and merge.

Great, sounds good!

PR #34, opened 2026-04-01
VS Code Releases

pixilcode:
Released VS Code versions `0.2.0` and `0.2.1`. `0.2.1` was released because I didn't test `0.2.0`.

I also added docs about publishing and testing the VS Code extension.


pixilcode:
Fuzz tests are failing, I will address that elsewhere.

PR #35, opened 2026-04-02
User docs

pixilcode:
Update the user docs and the README quick start.

Also has an unrelated addition of displaying test references in trees. I can move that commit to a seperate PR if you have a lot of questions about it. Basically, tests now are displayed as dependents alongside other parameters.


docs/guide/02-parameters.md
patrickwalton: I think an introduction with a mermaid diagram of how Oneil is structured from the user perspective would be helpful here. Could show a tree of a model with a few imported submodels and python. The model and submodels could show there are parameters, tests, derivation latex. The Python file could show a tiny sample of key syntax for interfacing. What do you think?

patrickwalton: You should also define what a parameter is. Something like "Parameters are variables with extra metadata for system modeling and review. Parameters define a long name, limits, a math symbol for rendering and review, units, a derivation, and either a value assignment or an equation relating this parameter to others."

patrickwalton: These terms are less clear about the purpose of those parts of the grammar.

Name is problematic, because many are going to think of the "human-readable name" as the name. I think few would associate the ID/symbol with a "name". I think either ID or maybe better yet, "symbol" would be a better term.

Label is confusing because people often associate labels with metadata, not the main name that a parameter would be called.

I find value misleading, because it implies a literal. I think a better term would be more general, either expression or assignment.

patrickwalton: Can you make a graphic here that shows big font parameter and then has the descriptions of each part of the syntax below?

[This](https://www.google.com/search?sca_esv=603be8ec33a5b75e&sxsrf=ANbL-n7M6meXl4drDMiyB3z8gcb33ojcWg:1775238460550&udm=2&fbs=ADc_l-aN0CWEZBOHjofHoaMMDiKpaEWjvZ2Py1XXV8d8KvlI3p-ML-906rRL_m6h4jR-tdCH-vUIlZq9RzugLEcfjf51avSTkNheog4cUvz8TkmXnk9I2ga_vexJisBM3m8mhi_5E3orE_jXYS7nyk5HDWiLRfoGIzmpYMi8bC4Pt9e4uVOPUNvlt4U1YVVFCL4bo6_HwGQ_ifCNBu5n6Q4SaTc1NmfhBg&q=parts+of+an+equation+image&sa=X&ved=2ahUKEwjJr9TOntKTAxWlIUQIHSA8NhkQtKgLegQIEBAB&biw=1600&bih=891&dpr=1.6#sv=CAMSVhoyKhBlLUJBUmNZU21EVU5VZ01NMg5CQVJjWVNtRFVOVWdNTToONDE1cUdEbjg3bFZmN00gBCocCgZtb3NhaWMSEGUtQkFSY1lTbURVTlVnTU0YADABGAcg4Y-f-wRKCBABGAEgASgB) is a rough idea, but of course a little too childish.

patrickwalton: When I wrote the first Oneil documentation, I spent some time try to write examples that would have character and start to spark ideas about what people could do with the language. In rewriting these. You have stripped the character out. 

Good writing is not ascetic. Good writing has personality and makes the reader feel like they are talking to a friend or at least a human. Part of why that's true is because the purpose of technical writing is utilitarian: get more readers to learn the subject matter. Dry content discourages readers.

See if you can go back and start to build something with your models. If you can update the Oneil cylinder examples, you should do so to save time and avoid reinventing. I like the idea of building an Oneil cylinder, since it fits with the name. In the future, we could do something a little more familiar to prospective users, maybe a rocket.

patrickwalton: Another example that should be building something (and they should all build the same thing progressively, as the original oneil docs did)

Can you add a task to the roadmap for example models? We should make those and then reference them in a new quickstart page of the guide.

patrickwalton: link annotations section again

patrickwalton: I think we should get rid of this header when using the CLI. The outputs are bulky because of it. I originally designed the header for the REPL.

We can instead warn the user about failing tests when the evaluate params, but only mention tests in the output when there are failing tests. Something like, "2/10 tests failed. Run `oneil <model name> tests` to see results"

patrickwalton: b: b=3 is redundant syntax. Why does it print this way?

patrickwalton: Used to identify or star the most important parameters in the model.

Side note, I like that you made it so oneil <model-name> defaults to eval, but we should also default to eval with "perf".

patrickwalton: Incomplete TODO. Is this still needed?

patrickwalton: Is this an example file name? Reading this I'm wondering if I'm supposed to put annotations in a different file called annot.on.

patrickwalton: I think this would be clearer if you were marking examples you'd already done:

```
# Trace marker:
* First: a = 1

# Debug marker:
** Second: b = a + c

# Performance marker:
$ Third: c = 2

Comment (patrickwalton) [reply docs/guide/02-parameters.md]: To do this with the comments, you'd need to introduce comments first.

Comment (patrickwalton) [reply docs/guide/02-parameters.md]: It might be good to put that graphic in an introduction that answers one question upfront: Why should I read these docs and use Oneil?

You could seed the introduction with the intro to the original docs.

**Review by patrickwalton — CHANGES_REQUESTED** [review]
First off, small thing, but can you install Cursor extension "markdownlint" and address all linter errors?

I have a lot of comments. I understand it will take you time to address them, but I still think it is good we pushed to have this done sooner. Some of these you may have addressed yourself with more time, but I suspect many you would not have. This way we end up with finished docs with feedback in the same time you might have finished without.

# Comments on your 01-installation.md
(since it doesn't show up in files changed)
- include the python-devel dependency next to the python dependency. (mention pip here as well instead of in later).
- Can you ask Ben to test with 0.5 on Windows? Unless you tested Windows yourself.
- Does install.sh install the python library? I think that would be a good thing. Then you could mention in the python library section
- Can you handle any of the troubleshooting steps in the install.sh? If we can do this and prev bullet, we should say that install.sh is the preferred install method and point the other methods to a section that fills in these gaps.

> Editable Python install (for working on the bindings)

I'm not quite sure what working on the bindings means or whether it applies to the average user.

> If Cargo is missing, the script prints how to install Rust via [rustup](https://rustup.rs/).

Not necessary in documentation. Users can cross that bridge when they get there.

> Vim syntax highlighting is not part of the script; see the README [Vim support](https://github.com/careweather/oneil#vim-support) section.

Premature reference. I would just remove. User hasn't encountered Vim yet.

> Vim: See the Vim support section in the main README for syntax highlighting.

if you haven't updated the vim syntax highlighter, you should mark this as not maintained and add a task to update with new grammar (maybe already is one)

# Other Comments
- Do we have a way to build the docs using our own syntax highlighting for code blocks? We currently have a couple domains for oneil (oneil-lang.org, oneil.systems), but I am also trying to get oneil.org. Either way, we could host the docs at the domain.

File: docs/guide/02-parameters.md [hunk 1]
```diff
@@ -210,8 +210,6 @@ p = 3  # Perf param
   p = 3  # Perf param
   ```
 
-- **`--debug` / `-D`** - When set, for parameters marked with `**`, Oneil prints debug information about the variables used to evaluate that parameter. Run `oneil eval --help` for current behavior.
-
 ## Other useful eval options
```
patrickwalton: In general, I feel like this chapter is getting too deep into the CLI causing it to be too long for an introduction to parameters. I think we should instead have a man page for the CLI and/or have a separate chapter.

docs/guide/03-tests.md
patrickwalton, line 1: I like how concise this one is, but it could maybe use a couple more examples.

I also think it's showing up too early. For example, the user hasn't been introduced to units yet.

docs/guide/04-value-types.md
patrickwalton: Terminology is definitely going to trip people up here.

When I read this line, my first question is what about equations? 

I would expect this intro to read something like:

"""
the parameter expression can include literals, equations, and imported functions.

### Literals
The main literal types are...
"""

patrickwalton: I hear fractional parts and expect to see a fraction (e.g. 1/2)

patrickwalton: This section and the next section are between what I'd expect for a language reference and a guide. You should simplify them to be more guide-like and save the detailed descriptions for the language reference.

patrickwalton: Users will assume they can do this, so the guide doesn't need this line. Good for the reference though.

docs/guide/04-value-types.md
```diff
@@ -20,7 +20,7 @@ Scientific: e_val = 1.5e3
 Infinity: inf_val = inf
 ```
 
-### Operations on numbers
+### Operations
```
patrickwalton: "Operators"

docs/guide/04-value-types.md
patrickwalton: This is too specific to remote sensing. We need another example case.

Also, labels is confusing when you used it as a name for part of the grammar earlier.

patrickwalton: I'm not sure what this means.

docs/guide/src/05-units.md
patrickwalton, line 1: This chapter should show up sooner. Content should be arranged from most relevant to the user / most frequently used to least.

docs/guide/05-units.md
patrickwalton: "One of Oneil's defining features is its unit-based type system. Oneil tracks units, disallows invalid operations between different physical properties, and automatically converts between differing units of the same physical property. For example, Oneil will throw an error if you try to add a time and a distance or compare a mass and a temperature. But it will automatically convert a length in meters and a length in feet to a common base before adding them together. This simplifies expressions to focus on relationships between physical properties while preventing unit conversion errors."

I mentioned ages ago that we should echo some of the arguments for built-in units that were given in the Rust units library you shared. Please do so.

patrickwalton: The style guide is no space between colon and unit. Use that for the documentation.

patrickwalton: This is cool. Is the search exact match or fuzzy?

patrickwalton: Update examples for limited sifigs. When I see a decimal with lots of zeroes, my mind tries to read it as a  comma for a very large number.

patrickwalton: Again the terminology here is making the error message less helpful. The original error was more helpful. Something along the lines of "calculated unit does not match expected unit"

With this, your text should be clearer about the varying role the unit plays from the user perspective, for an independent parameter, the unit defines the physical property assigned to a literal. For a calculated parameter, the physical property assigns the expected physical property of the calculation.

patrickwalton: This doesn't make sense. You should say "any parameters whose calculations are expected to produce a measured value" or something like that.

patrickwalton: "Likewise, the calculation for a unitless parameter should not have a measured result. In that case, we do leave the units out (see dimensionless values).

patrickwalton: Not sure this is correct markdown. Just renders as "![NOTE]"

patrickwalton: "dimensions, listed here with their..."

patrickwalton: This is better writing, but really should show up at the top of the chapter to fill additional purposes:
- explain valid operations (does this already)
- set terminology dimensions, units, magnitude, etc. (we then need to use that terminology consistently for the rest of the chapter)
- explain the motivation (the rust units library does a great job of this). No explanation is complete without a reference to Mars Climate Orbiter.

patrickwalton, line 61: "Base units convey a single dimension, like the unit `kilometer` with its dimension `distance`. Derived units convey anywhere from 0 to many dimensions, like the unit `degree` which is dimensionless or the unit `Joule` with its dimensions of of `mass`, `distance^2`, and `time^-2`. Dimensionless units are discussed in more detail [later in this chapter](#unitless-and-dimensionless-values)."

patrickwalton: Generally, you can get away with more details earlier on if they are engaging. I think this text is engaging, in part because the writing is good and in part because it is not syntax heavy.

patrickwalton: I have to know: did AI write this line?

"Delve" has become something of a tracer for AI-generated text. AIs seem to like the word a lot.

patrickwalton: This use of "base unit" does not align with SI definitions: https://en.wikipedia.org/wiki/List_of_metric_units

I think I understand what you're trying to say, but I don't think you need to say it. The text above is clear enough. I think the users understand that a unit expression can be build from "one or more units" though you might use that phrase at the intro to this section.

patrickwalton: "are defined"

patrickwalton: Table isn't rendering correctly...

patrickwalton: I'm not sure "unitless values have a unit of 1" is helpful to the reader.

patrickwalton: I think you mean "b" cannot have a dimension? If not, we should add plans to the roadmap to handle [this](https://www.google.com/search?q=examples+of+equations+where+something+is+raised+to+radians&oq=examples+of+equations+where+something+is+raised+to+radians&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIHCAEQIRigATIHCAIQIRigATIHCAMQIRifBTIHCAQQIRifBTIHCAUQIRifBTIHCAYQIRifBTIHCAcQIRifBTIHCAgQIRifBTIHCAkQIRifBdIBCTEyODA5ajBqN6gCALACAA&sourceid=chrome&ie=UTF-8).

patrickwalton: typo

patrickwalton: You haven't introduced unit casting, nor (after reorganization) will you have introduced tests. As a result, these examples are going to be lost on the reader.

patrickwalton: You should come up with a less-contrived situation then.

docs/guide/05-units.md
```diff
@@ -290,7 +290,7 @@ The places where this should be used are rare and should be treated cautiously
 since `strip` effectively disables unit checking.
 
 ```oneil
-X: x = 10 :m
+X: x = 10 : m
 
 test: strip(x) == 10
```
patrickwalton: show the results. It might be helpful to write these examples and their results in the format the user will see when the repl is done.

docs/guide/05-units.md
patrickwalton, line 405: I don't get this. Why wouldn't I just use a literal?

docs/guide/07-notes.md
patrickwalton, line 1: We should either remove these files, or write the bare minimum guide for them. What is the quickstart for each of these concepts?

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
 ```
 
 
-## Unitless and dimensionless values
+## Dimensionless units
 
-TODO
+There are some units that don't have any dimensions, such as `%` or `ppm` (parts
+per million). The plain `dB` unit also does not have any dimensions.  These
```
patrickwalton: We are going to need a section on nonlinear units. We should also add a task to the roadmap to flesh out units further, including a variety of [other nonlinear units](https://www.google.com/search?q=examples+of+equations+where+something+is+raised+to+radians&oq=examples+of+equations+where+something+is+raised+to+radians&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIHCAEQIRigATIHCAIQIRigATIHCAMQIRifBTIHCAQQIRifBTIHCAUQIRifBTIHCAYQIRifBTIHCAcQIRifBTIHCAgQIRifBTIHCAkQIRifBdIBCTEyODA5ajBqN6gCALACAA&sourceid=chrome&ie=UTF-8).

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
...
+units are referred to as _dimensionless units_, and values with dimensionless
+units are referred to as _dimensionless values_.
+
+### Unitless equivalence
+
+Dimensionless values are treated as if they have no unit for all purposes except for
```
patrickwalton: I had to read this three times to understand what you were trying to say. It will get clearer when we move dB examples into a nonlinear section, but you need to try again at writing this one clearer.

patrickwalton: I think part of the problem is I'm not sure if you defined these terms anywhere "dimensionless vs. unitless". If you did, you maybe need to repeat the definition.

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
...
+
+test: (360:deg) == 2*pi
+
+test: (360:deg) + 2*pi == (720:deg)
+```
+
+#### Using `rad/s` and `Hz`
+
+In terms of SI units, there is a problem with `rad/s` and `Hz`. It is that
+[`rad/s != Hz` even though `rad/s == 1/s` and `Hz == 1/s`](https://iopscience.iop.org/article/10.1088/1681-7575/ac0240).
```
patrickwalton: no code snippets in links, markdown doesn't handle them well

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
...
+In order to get around this problem, Oneil specifies that `1 Hz == 2*pi rad/s`.
+This solution helps us to avoid this inconsistency, but it can also lead to
+some problems of its own.
+
+```oneil
+Frequency: f = 1 : Hz
+Cycles in 2 seconds: cycles = f * (2:s) : 1
+# we would expect `2`, since 1 Hz = 1 cycle/s
```
patrickwalton: Above, you said "1 Hz == 1/s" but here you contradict that.

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
...
+
+The `cycle` unit is defined as `2*pi radians`, so if you simply want to view the
+unit in cycles, add the `cycle` unit.
+
+```oneil
+Cycles in 2 seconds: cycles = f * (2:s) : cycles
+# prints out `2 cycles`
+```
+
+However, using the `cycle` unit on an intermediate value will cause `cycle` units
```
patrickwalton: I think the cycles discussion is overall distracting here. We should say something like, "there is one case where you still need to do unit conversions" then explain how to handle this problem.

docs/guide/04-units.md
```diff
@@ -290,7 +290,7 @@ The places where this should be used are rare and should be treated cautiously
 since `strip` effectively disables unit checking.
 
 ```oneil
-X: x = 10 :m
```
patrickwalton, line 293: A great example use case for strip shows up in radar.on where you want to calculate a range of possible numbers using n bits, but we have bits as a unit, so 2^n is not valid.

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
...
+Wavelength: lambda_convert2 = 2*pi*c/f : cm
+```
+
+In the future, we may add a rotation dimension, but for now, this is what is
+required to get around the inconsistency.
+
+
+### Display
+
+As mentioned previously, dimensionless units are mainly kept around so that
```
patrickwalton: It's been so long, I don't even remember you mentioning previously. Just remove.

Also, you should just say "Dimensionless units allow the user to specify values and see results in a more human-readable form. For example..."

docs/guide/05-units.md
```diff
@@ -318,6 +318,168 @@ test: strip((x_m : m)) == strip((x_km : m))
...
+
+```oneil
+Length 1: l1 = 10 : m
+Length 2: l2 = 20 : m
+
+$ Length ratio: lr = l1 / l2
+# prints out as `0.5`
+```
+
+There are some cases, though, where Oneil is unable to detect that a value is
```
patrickwalton: This is a bug and the user shouldn't be asked to work around it. No ratio should ever default to km/m. The default should be just 0.5, unless the user specifies percentages or something else. We need to fix this before we release.

docs/guide/06-intervals.md
```diff
@@ -0,0 +1,110 @@
...
+```oneil
+X: x = 10 | 15
+Y: y = 0 | 5
+Z: z = x - y
+# => (10 | 15) - (0 | 5)
+# => 10 - 5 | 15 - 0
+# => 5 | 15
+```
+
+All arithmetic operators are defined to give correct interval bounds. For more
```
patrickwalton: Delete first sentence.

docs/guide/06-intervals.md
```diff
@@ -0,0 +1,110 @@
...
+[interval arithmetic paper review](../../research/2025-11-13-interval-arithmetic-paper-review.md)
+or the implementation in the codebase.
+
+### Escaping and relationships
+
+Oneil’s interval arithmetic aims to satisfy the *inclusion property*: if every
+interval in an expression is replaced by some scalar inside that interval and the
+expression is evaluated as scalars, the scalar result lies inside the interval
+you get by evaluating the expression on intervals. See
+[Inclusion property](../../research/2025-11-13-interval-arithmetic-paper-review.md#inclusion-property)
```
patrickwalton: Just hyperlink inclusion property at top of paragraph and remove this last sentence.

patrickwalton (commented):
Sorry, I hit submit too early. Generally, I wanted to add that this is a really good start. Thank you.

# Other comments
- The guide (in contrast to a language reference), should be written as though you're teaching a friend to use the language. One exercise you can use to see how far away you are is to rewrite paragraphs strictly in second-person. "To write a parameter, you can do x, y, z."

docs/guide/07-notes.md
patrickwalton: Wrote a notes chapter and pushed it.

docs/guide/02-parameters.md
pixilcode: > I think an introduction with a mermaid diagram...

I'm not sure I understand exactly what you're looking for with the diagram, but I can definitely add one! Maybe you can sketch what you're envisioning and send it to me?

- [x] Add mermaid diagram
- [x] Define what a parameter is

docs/guide/05-units.md
pixilcode: No, the units chapter was mostly written by me. :sweat_smile: I can change the word "delve" if you'd like though.

pixilcode:
> Do we have a way to build the docs using our own syntax highlighting for code blocks? We currently have a couple domains for oneil (oneil-lang.org, oneil.systems), but I am also trying to get oneil.org. Either way, we could host the docs at the domain.

We don't currently have a way, but I don't think it would be hard to implement one. There are plenty of syntax highlighting libraries out there (see [highlight.js](https://highlightjs.org/), [bright](https://bright.codehike.org/), and [pygments](https://pygments.org/)). All we'd have to do is describe the grammar in a way that the library understands and it will fill in the rest.

docs/guide/02-parameters.md
pixilcode: - [x] Parameter graphic
- [x] Introduction: Why should I read these docs and use Oneil?

pixilcode: - [x] Models should have character

pixilcode: - [x] Update CLI output

pixilcode: It's because I was reusing a piece of code to print the parameter. I've updated the CLI to remove the extra name.

- [x] Update CLI

pixilcode: No, I will delete it.

- [x] Delete TODO

pixilcode: - [x] CLI defaults to `perf` instead of `trace`

pixilcode: - [x] Fix annotations example file names

docs/guide/02-parameters.md
pixilcode: - [x] Remove CLI description from parameters chapter

docs/guide/04-value-types.md
pixilcode: - [x] Improve Value Types description

pixilcode: - [x] Replace with "decimal parts"

pixilcode: Is this section and next section referring to `## Numbers` and `### Number Operations` or `## Numbers` and `## Strings`, or something else?

pixilcode: - [x] Remove the "optional leading sign" item

docs/guide/04-value-types.md
pixilcode: - [x] Replace "Operations" with "Operators"

docs/guide/04-value-types.md
pixilcode: - [x] Find additional strings example
- [x] Find better wording than "labels"

pixilcode: - [x] Remove mention of operator binding

docs/guide/05-units.md
pixilcode: - [x] Include arguments for units from `uom`

pixilcode: Okay, I couldn't remember what the preferred style for that was (I saw both when I was reviewing the Veery code).

Where is the style guide found?

- [x] Remove spaces between colon and unit

pixilcode: Currently, it's exact match. I have code in there for fuzzy searching in other parts of Oneil though, I just haven't had a chance to incorporate it into the builtins search yet. Would you like me to do that?

pixilcode: - [x] Update sig figs in docs

pixilcode: - [x] Update error message wording
- [x] Make role of unit clear

pixilcode: - [x] Reword

pixilcode: - [x] Reword

pixilcode: - [x] Change `![NOTE]` to `[!NOTE]`

pixilcode: - [x] Reword

pixilcode: - [x] Move to top of chapter
- [x] Define terminology
- [x] Explain motivation (see `uom` crate)

pixilcode: - [x] Reword

pixilcode: - [x] Delete ponder note

pixilcode: - [x] Reword

pixilcode: - [x] Reword

pixilcode: - [x] Fix table rendering

pixilcode: - [x] Remove wording

pixilcode: - [x] "unit" -> "dimension"

pixilcode: - [x] Fix typo

docs/guide/05-units.md
pixilcode: The link goes to a Google search of "examples of equations where something is raised to radians", which you also linked to in a previous comment. Is that search related to non-linear units as well, or was there another link you wanted me to look at?

docs/guide/05-units.md
pixilcode: - [x] Write this section more clearly

docs/guide/05-units.md
pixilcode: - [x] Remove code snippets from link

docs/guide/04-units.md
pixilcode: - [x] Add strip example with bits

docs/guide/05-units.md
pixilcode: - [x] Clean up and clarify

docs/guide/05-units.md
pixilcode: Yes, I agree. At the time of writing, I wasn't able to find a solution to the bug, and I had already spent a considerable amount of time working on it, so I just let it go since it doesn't show up very often and only impacts how the unit is displayed, not how the calculations work. I included it in the guide in case anyone runs into it.

However, thinking over it the last two weeks, I've come up with a potential solution for it that may not need the work around. So I'll try implementing that and report back.

- [x] Fix printing bug

docs/guide/06-intervals.md
pixilcode: - [x] Delete first sentence

docs/guide/06-intervals.md
pixilcode: - [x] Move link

docs/guide/02-parameters.md
pixilcode: - [x] Move comments section up
- [x] Mark previous example rather than creating new example

docs/guide/03-tests.md
pixilcode: You're right, I forgot that units hadn't been introduced yet. I introduced tests early so that they could be used in later chapters without confusing the user.

- [x] Add more examples

docs/guide/src/05-units.md
pixilcode: Okay. I was trying to organize in a way that builds up from the basics.

- **Parameters** - parameters are the fundamental feature of models
- **Tests** - tests are another feature of models; this is mainly here so that I can use tests in the example code in future chapters to more clearly express facts
- **Value types** - parameter expressions can have different types, including numbers, strings, and booleans
- **Units** - numbers can have a special property known as a "unit"
- **Intervals** - numbers can also have a "min" and "max" value
- **Notes** - returning to parameters, parameters can have documentation
- **Importing Models** - moving from a single model to multiple models - a model can be imported by other models
- **Importing Python** - adding external functionality not supported by Oneil

 What do you consider the order of relevance to be in terms of chapters?

docs/guide/05-units.md
pixilcode: That's true. I can't think of a better way to do it. Do you have any ideas for how I would do it, or do you think I should get rid of the examples?

pixilcode: I can't think of one since I haven't encountered it, and I imagine that using it in a non-contrived situation would require a more complicated situation. I think that knowing about this use of unit casting may be useful even when demonstrated with a contrived example, but I can get rid of the `Verifying units` section if you'd like.

pixilcode: I don't think I understand your question. Where would you use a literal?

docs/guide/05-units.md
pixilcode: Are you talking about all the examples in the guide, this section, or just this example?

docs/guide/07-notes.md
pixilcode: - [x] Decide what to do for each concept

docs/guide/05-units.md
pixilcode: Are you talking about when I say that "`rad/s != Hz` even though `rad/s == 1/s` and `Hz == 1/s`"?

docs/guide/05-units.md
pixilcode: - [x] Rewrite

docs/guide/02-parameters.md
pixilcode: - [x] Rename

pixilcode: - [x] Add task

pixilcode: - [x] Add link

pixilcode:
I've gone through and added tasks to each piece of feedback. If you have any updates to the tasks feel free to comment them.

I've also marked comments that contain questions/need feedback with ' :confused: '.

docs/guide/04-value-types.md
pixilcode: I could use the example case of a battery's array configuration being either `'series'` or `'parallel'`. Would that be a good additional example case?

docs/guide/02-parameters.md
patrickwalton: Something like this: https://docs.google.com/drawings/d/1Nd-BlFXpJkyz85Us0s726nU97hpOT9fKmEJmg85k7Yc/edit

docs/guide/04-value-types.md
patrickwalton: "## Numbers and ### Number Operations"

patrickwalton: Yeah, that's more generally understood.

docs/guide/src/05-units.md
patrickwalton: Well, I would like us to have both a guide and a reference, but as I read what your writing, it seems clear you have a reference. In which case I would just move tests after units.

In the future we should add an introductory guide (quick start?) that builds up from the simplest to more complex code:
- unitless independent parameter (numeric literal, string literal), with all parts of the simplest parameter: name, symbol, value
- above but with limits
- above with a note
- above with units
- above with value interval
- dependent parameter (simple expression)
- dependent parameter (two-equation interval, mixed equation/literal interval)
- tests
- importing python
- model imports
- designs

docs/guide/05-units.md
patrickwalton: We don't have a style guide, but I was referring to what it would say if we had one.

I brain-dumped a lot of style into my oneil cursor rules which I think I pointed you to previously.

If we don't already have a plan for it, we should make sure the roadmap somewhere captures that we need to write a formal style guide and accompanying features in the linter.

patrickwalton: maybe add to roadmap for later (can just be in a work plan or issue)

patrickwalton: Haha, no it's fine, I was just curious.

patrickwalton (commented):
Discarded Review (Github doesn't have a way to cancel a review)

docs/guide/05-units.md
patrickwalton: Well, this is the units chapter, so you could just move unit casting up.

For just showing comparison, you should introduce the concept of an expression in Oneil early on (parameters or introduction) and just have a way to show an expression without putting it in a test.

docs/guide/05-units.md
patrickwalton: In the guide in general.

docs/guide/05-units.md
patrickwalton: I think I understand now and the confusion was caused by not showing the results of your example expressions.

patrickwalton: I'm still unclear on what you're trying to say here? Are you saying that once I try and cast the unit, then I'll see the error for v_start being the wrong unit? why didn't I see the error before? This might be an issue stemming from not showing the user the results for different calculations. Above, you say "you might be trying to debug the following model, which has a unit error", but you don't tell me what the error was! So you're giving me an example that highlights the error, but then you leave me to guess at how your example is changing the errors.

docs/guide/05-units.md
patrickwalton: correct link: https://www.google.com/search?q=examples+of+nonlinear+units&oq=examples+of+nonlinear+units&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIHCAEQIRigATIHCAIQIRigATIHCAMQIRigATIHCAQQIRigAdIBCDUzNzJqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8

docs/guide/05-units.md
patrickwalton: Move your definition of the cycle unit above this example and I think it will be clear.

Also the :1 makes that line extra confusing.

patrickwalton:
K, I believe I've addressed all of your questions.

docs/guide/05-units.md
pixilcode: Okay. I agree that I'm not explaining it well enough. I think that it's probably not useful enough for me to try and rewrite, and the explanation is too confusing to keep, so I'll just get rid of the section. My time would probably be better spent elsewhere.

- [x] Delete section

docs/guide/05-units.md
pixilcode: - [x] Add outputs throughout the guide

docs/guide/05-units.md
pixilcode: - [x] Add roadmap task

docs/guide/04-value-types.md
pixilcode: - [x] Simplify numbers section

docs/guide/src/05-units.md
pixilcode: - [x] Add overview section

docs/guide/05-units.md
pixilcode: - [x] Add style guide + linter to roadmap

pixilcode: - [x] Add to roadmap/issue

pixilcode: - [x] Move unit casting up
- [x] Use `-x` flag for examples rather than tests

docs/guide/05-units.md
pixilcode: - [x] define cycle unit above example
- [x] remove `:1`

docs/guide/src/10-designs.md
```diff
@@ -0,0 +1,352 @@
...
+submodel spacecraft as vehicle
+
+Mission duration: t_mission = 365 :days
+```
+
+```oneil
+# boosted_mission.one
+design mission
+
+thrust.vehicle.main_engine = 2000 :N
```
TimWhiting: @patrickwalton 

This is what I was talking about. We are designing mission, it has a spacecraft submodel we call vehicle, and that vehicle has a main_engine which has a thrust. 

Alternatives:

1. `thrust.vehicle.main_engine` is weird because it is the thrust of the main_engine, not the whole vehicle.
2. `thrust.main_engine.vehicle` seems strange at least to programmers and english readers since to verify that the parameter is accessible you have to read from right to left to make sure you used valid submodel names / references. 

At one point early on I think we discussed that we should disallow nested `.` even for designs, but later I remember you wanting it. Since we can't add `[main_engine]` to the submodel in mission from the design file, we don't currently have any other way of modifying thrust other than creating a design for spacecraft and applying it to vehicle (alternative 3). 

In other words, we either need to provide a justification for one of the first two alternatives or make clear that nested access to parameters is disallowed and applying designs should be used instead. 

I remember you saying that we should support nesting because designs are too much overhead for a single parameter change.

patrickwalton:
Github is giving me an error when I try to review, so here are my comments:

> oneil mars_gravity.one -P all

Curious why this doesn't require `eval` and if this is allowed, can we run basic oneil files without eval?

```
# Each planet is an independent instance.
submodel planet as earth
submodel planet as mars
apply mars_gravity to mars
```

Each example should also show the user how to assign symbols to imported models.

```
Since `constants` is a `reference`, if you changed `G` with a design, both
`earth` and `mars` would see the update — which is exactly what you want for
shared environmental data. For the planets themselves, `submodel` isolation
means you can give each one a different design independently.
```

These needs a new example since no one is going to understand why we would be re-designing the universal gravitational constant. I've re-read this many times and I'm still also unclear on what "changing G with a design" would look like. Can we add a code block demonstrating this?

General comment: is there any value to manually line breaking the text in the guides? I find it tedious to maintain and unnecessary now that modern editors let you wrap text like VS Code with CTRL + Z.

> ### Sibling designs

Concepts from this section get used in the previous one. Move it up.

> You can override a parameter on a submodel directly from a design using dotted syntax: parameter.submodel = value.

There are so many files to this example I'm getting lost. If we pick a common example to build up over the course of the chapter, then you don't have to repeat all the files, you can just imply/import a previous file and the user will naturally see where things are going.

> submodel thruster as main_engine

> thrust.main_engine.vehicle seems strange at least to...english readers

I think we need to decide if we are going to design to support long names or not for one thing. I don't think that is our intent. Part of the purpose here was to give people the option to use symbols and I think we can be opinionated about that.

But to your question about nesting, I don't think we should allow it. Anything beyond a direct import alias should have to be aliased as a direct import. Does that make sense. In other words, none of the main models are going to allow you to use a parameter from the submodel of a submodel unless you first import that submodel of a submodel as an alias. The designs should follow the same constraint.

> Adding performance and other parameters

The examples confuse me here and in the following section too. At some point we'll need you to build a large model from scratch. That will help you come up with better examples. Since most of the document is the examples (which is good!) we should invest in them more.

patrickwalton:
Realized there were a few other chapters I haven't reviewed yet. 

But first off, the writing is generally looking good! I think the biggest problem we have is that you guys aren't users of Oneil, so we're stuck on examples that aren't terribly illustrative. In general, I don't like any of the planet examples, we need users to immediately get ideas of what cool things they could model quickly with this rather than just talking about planets which already exist.

Looking at chapter 9:

> Both photon.on and link_budget.on read from the same constants.on file. If a [design](https://file+.vscode-resource.vscode-cdn.net/home/patrickwalton/careweather/oneil/docs/guide/src/10-designs.md) overrides a value on constants, both models see the change — that's what it means to share a reference.

Premature reference (unless you give a short introduction to designs in an earlier chapter). I think you can just introduce concept in designs, but also I've read about this multiple times and am still not quite understanding it.

> The as alias is optional — without it, you access parameters using the model name:

Are we overloading `as` seems like here its giving an alias but in designs [] gives the alias or is that only aliases for submodels?

```
submodel planet as mars
apply mars_design to mars
```

Can we do this in one line? Maybe `apply mars_design to submodel planet as mars`?

```
To access a parameter inside a reference or submodel, write `parameter_id.alias`
— the **parameter comes first**, the model second.

oneil
# satellite.on
reference constants as phys
submodel planet as target

Orbital speed: v = sqrt(G.phys * M.target / R.target) :m/s
```

The primary example here should use symbols, not long aliases.

> When a submodel itself contains submodels, you can declare a local alias for one of those inner models using [alias] at the end of the submodel line — giving you direct access to a nested model.

Change to "you must declare...to access the parameters of those inner models"

```
# galaxy.on
# Import the solar system and declare `earth` as a local alias here,
# so it can be read and overlaid directly at the galaxy level.
submodel solar_system as sol [earth]
```

I think this is what I'm confused about. I think maybe we need to give `[earth as e]` here and then let the user infer from the previous comment about leaving out alias / symbols that we can do that in subsubs as well.

Generally, we need to be clearer that one of the main values of aliases are for symbol usage, otherwise, people are not going to get the reason they're in here.

Now moving to chp 11:

> For functions not supported by the above equation formats, you can define a

"above"? I'm not sure what that is referencing.

> We don't currently have a way, but I don't think it would be hard to implement one. There are plenty of syntax highlighting libraries out there (see [highlight.js](https://highlightjs.org/), [bright](https://bright.codehike.org/), and [pygments](https://pygments.org/)). All we'd have to do is describe the grammar in a way that the library understands and it will fill in the rest.

Also, just want to comment that this would be very helpful.

pixilcode:
> >We don't currently have a way, but I don't think it would be hard to implement one. There are plenty of syntax highlighting libraries out there (see [highlight.js](https://highlightjs.org/), [bright](https://bright.codehike.org/), and [pygments](https://pygments.org/)). All we'd have to do is describe the grammar in a way that the library understands and it will fill in the rest.
>
> Also, just want to comment that this would be very helpful.

This is done! :grin:  For a good example, see [the user guide](https://careweather.github.io/oneil/#model-imports).

TimWhiting:
> Github is giving me an error when I try to review, so here are my comments:
> 
> > oneil mars_gravity.one -P all
> 
> Curious why this doesn't require `eval` and if this is allowed, can we run basic oneil files without eval?

Yes you can run without explicit eval command

> ```
> # Each planet is an independent instance.
> submodel planet as earth
> submodel planet as mars
> apply mars_gravity to mars
> ```
> 
> Each example should also show the user how to assign symbols to imported models.

This does, `earth` / `mars` are the symbols referring to each planet model. 
Are you expecting `submodel planet as m # mars`? 

> 
> ```
> Since `constants` is a `reference`, if you changed `G` with a design, both
> `earth` and `mars` would see the update — which is exactly what you want for
> shared environmental data. For the planets themselves, `submodel` isolation
> means you can give each one a different design independently.
> ```
> 
> These needs a new example since no one is going to understand why we would be re-designing the universal gravitational constant. I've re-read this many times and I'm still also unclear on what "changing G with a design" would look like. Can we add a code block demonstrating this?

Probably the better example for your reference is the altitude parameter of the balloon test. Something that multiple models want to refer to, but designed such that every model depending on it sees the updated value. 

> General comment: is there any value to manually line breaking the text in the guides? I find it tedious to maintain and unnecessary now that modern editors let you wrap text like VS Code with CTRL + Z.

I find it easier to read especially if the screen is wide, however, I think a good compromise is to break on periods.


> > ### Sibling designs
> 
> Concepts from this section get used in the previous one. Move it up.

Sounds good

> > You can override a parameter on a submodel directly from a design using dotted syntax: parameter.submodel = value.
> 
> There are so many files to this example I'm getting lost. If we pick a common example to build up over the course of the chapter, then you don't have to repeat all the files, you can just imply/import a previous file and the user will naturally see where things are going.

Good point.

> > submodel thruster as main_engine
> 
> > thrust.main_engine.vehicle seems strange at least to...english readers
> 
> I think we need to decide if we are going to design to support long names or not for one thing. I don't think that is our intent. Part of the purpose here was to give people the option to use symbols and I think we can be opinionated about that.

I think it is important to be able to use long names. You only have 26 letters of the alphabet (plus I guess unicode) if you want to use a single symbol, unicode can be hard to type and if `m` is already used for `mars` then if you need it for something else that starts with `m` that could be unfortunate.

> But to your question about nesting, I don't think we should allow it. Anything beyond a direct import alias should have to be aliased as a direct import. Does that make sense. In other words, none of the main models are going to allow you to use a parameter from the submodel of a submodel unless you first import that submodel of a submodel as an alias. The designs should follow the same constraint.

I like this opinionated view. I will remove support.

> > Adding performance and other parameters
> 
> The examples confuse me here and in the following section too. At some point we'll need you to build a large model from scratch. That will help you come up with better examples. Since most of the document is the examples (which is good!) we should invest in them more.

I will improve the examples with this feedback and the other feedback below.

TimWhiting:
> Realized there were a few other chapters I haven't reviewed yet.
> 
> But first off, the writing is generally looking good! I think the biggest problem we have is that you guys aren't users of Oneil, so we're stuck on examples that aren't terribly illustrative. In general, I don't like any of the planet examples, we need users to immediately get ideas of what cool things they could model quickly with this rather than just talking about planets which already exist.

To me modeling planets seems natural when talking about orbital mechanics / planetary missions. The examples are simplified to show the structure and highlight the particular feature we are introducing, but I could (1) improve those to use the planet's gravity to calculate escape velocity or something, or (2) spend time looking at the current oneil models and come up with something completely different. I guess though these don't really showcase the typical usage of designs. 

> Looking at chapter 9:
> 
> > Both photon.on and link_budget.on read from the same constants.on file. If a [design](https://file+.vscode-resource.vscode-cdn.net/home/patrickwalton/careweather/oneil/docs/guide/src/10-designs.md) overrides a value on constants, both models see the change — that's what it means to share a reference.
> 
> Premature reference (unless you give a short introduction to designs in an earlier chapter). I think you can just introduce concept in designs, but also I've read about this multiple times and am still not quite understanding it.

Okay, I think it is fine to remove this reference to designs

> > The as alias is optional — without it, you access parameters using the model name:
> 
> Are we overloading `as` seems like here its giving an alias but in designs [] gives the alias or is that only aliases for submodels?

`[]` just gives aliases for submodels. Aliases are always after `as`. 
So `submodel engine as e [thrust as t]` should be clear -> `e` and `t` are aliases to the child (engine) and grandchild (thrust) models.

The `[thrust as t]` though aliases a submodel or alias. If `t` was what thrust was called in the engine model we could call say `[t as t1]` here.



> ```
> submodel planet as mars
> apply mars_design to mars
> ```
> 
> Can we do this in one line? Maybe `apply mars_design to submodel planet as mars`?

Happy to do this. Simple syntax addition.

> ```
> To access a parameter inside a reference or submodel, write `parameter_id.alias`
> — the **parameter comes first**, the model second.
> 
> oneil
> # satellite.on
> reference constants as phys
> submodel planet as target
> 
> Orbital speed: v = sqrt(G.phys * M.target / R.target) :m/s
> ```
> 
> The primary example here should use symbols, not long aliases.

Sounds good will change to `p` and `t`. That is unless I change up all the planet examples to something else.

> > When a submodel itself contains submodels, you can declare a local alias for one of those inner models using [alias] at the end of the submodel line — giving you direct access to a nested model.
> 
> Change to "you must declare...to access the parameters of those inner models"

Thanks 


> ```
> # galaxy.on
> # Import the solar system and declare `earth` as a local alias here,
> # so it can be read and overlaid directly at the galaxy level.
> submodel solar_system as sol [earth]
> ```
> 
> I think this is what I'm confused about. I think maybe we need to give `[earth as e]` here and then let the user infer from the previous comment about leaving out alias / symbols that we can do that in subsubs as well.
>
> Generally, we need to be clearer that one of the main values of aliases are for symbol usage, otherwise, people are not going to get the reason they're in here.

Earth as e sounds good. In other words the main value of `alias`es is not to distinguish the planet model for earth from the planet model for mars, but to have short "subscripts".

patrickwalton:
> To me modeling planets seems natural when talking about orbital mechanics / planetary missions.

I think you should mostly leave the examples as is for now, with another pass for clarity. You will get much better intuition as you build some large models yourself.

PR #36, opened 2026-04-02
Fallback support

pixilcode:
This adds the fallback operator (`?`) to Oneil, allowing users to define fallback equations when Python functions throw an error.

_Caching is not completed yet. That will come in a seperate PR. We might as well get this part out there for use since it's done, though._


TimWhiting:
I guess I reviewed this at the same time as the other since that builds on this. Comments there are applicable here as well.

TimWhiting:
Once the CI checks are fixed, this looks good to me.

patrickwalton:
@pixilcode Can you add to your description either a summary of the fallback logic or point me to docs that describe it as intended so I can review at the user level?

pixilcode:
Yep! There's docs in the user guide [in the overview](https://github.com/careweather/oneil/blob/user-docs/docs/guide/src/01-overview.md#fallback-operator) and [in the python imports chapter](https://github.com/careweather/oneil/blob/user-docs/docs/guide/src/11-importing-python.md#fallback-calculations).

The underlying algorithm can be summarized as:

1. Given a "fallback expression" `E1 ? E2`, evaluate `E1`.
2. If `E1` succeeds, return the result.
3. If `E1` fails, then check the result errors.
4. If there are any errors that are **not** Python evaluation errors, return those errors
   - _All Python evaluation errors are removed from the list of errors that is returned_
5. If all errors **are** Python evaluation errors, emit a warning, then evaluate `E2` and return the result.

Does that make sense, or would you like me to go into more detail?

patrickwalton:
No that sounds good. Feel free to merge and have tim approve if approval is needed still

Take care,

Patrick Walton
Co-founder/CEO
Care Weather
801-227-4740

pixilcode:
The failing snapshot test found a bug in error printing and spans that will be handled in a seperate PR.

PR #37, opened 2026-04-03
Install updates

pixilcode:
Fix the linux install script and align the windows install script with the linux script.


pixilcode:
Fuzz test is failing because of interval arithmetic problems. This will be fixed in a later PR.

PR #41, opened 2026-04-30
Add Python import tracking

pixilcode:
_NOTE: This PR is dependent on [Fallback support](https://github.com/careweather/oneil/pull/36) and should not be reviewed until that PR is merged_

This PR adds tracking for Python imports within a script by replacing the standard import mechanism (`builtins.__import__`) with a custom mechanism that tracks local imports.

### DIFF
--- a/src-rs/oneil_cli/src/lib.rs
+++ b/src-rs/oneil_cli/src/lib.rs
@@ -12,6 +12,8 @@ use anstream::{ColorChoice, eprintln, print, println};
 use clap::Parser;
 use indexmap::{IndexMap, IndexSet};
 use notify::Watcher;
+#[cfg(feature = "python")]
+use oneil_runtime::output::PythonModule;
 use oneil_runtime::{
     Runtime,
     output::{
@@ -146,19 +148,15 @@ fn apply_common_side_effects(common_args: &CommonArgs) {
 fn handle_print_python_imports(files: &[PythonPath], show_internal_errors: bool) {
     let mut runtime = Runtime::new();
 
-    let mut imports: IndexMap<PythonPath, IndexSet<String>> = IndexMap::new();
+    let mut imports: IndexMap<PythonPath, PythonModule> = IndexMap::new();
     let mut errors = Vec::new();
 
     for file in files {
         let import_result = runtime.load_python_import(file);
 
         match import_result {
-            Ok(import) => {
-                let functions: IndexSet<String> = import
-                    .into_iter()
-                    .map(|name| name.as_str().to_string())
-                    .collect();
-                imports.insert(file.clone(), functions);
+            Ok(module) => {
+                imports.insert(file.clone(), module.clone());
             }
 
             Err(runtime_errors) => {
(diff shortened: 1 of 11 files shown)


src-rs/oneil_cli/src/lib.rs
```diff
@@ -429,19 +461,23 @@ fn eval_and_print_model(
     let (result, model_errors, expr_errors) =
         runtime.eval_model_and_expressions(file, eval_expressions);
 
+    let mut model_saw_error_diagnostic = false;
```
TimWhiting: This pattern is happening a lot, can we extract the error handling into a helper function?

src-rs/oneil_output/src/error.rs
TimWhiting, line 1: Did you split this file into smaller ones on purpose? Any reason for doing so? I would prefer to see fewer changes in a PR unless they are truly needed.

README.md
```diff
@@ -511,6 +511,23 @@ In the Oneil file, give the python function on the right hand of the equation, i
 Temperature: T = temperature(D) :K
 ```
 
+##### Fallback Calculations
+
+Python functions may have dependencies that aren't always available, or may take a long time to run. You can specify a fallback calculation using the `?` operator. If the Python function fails (e.g., missing dependencies, runtime errors), Oneil will use the fallback and warn the user:
```
TimWhiting: We should be clear that timeout is not currently supported, or support it with a flag.

TimWhiting (commented):
Looks good to me other than my bits of feedback. A lot of churn with the diagnostic change, which might have been better as a separate PR.

src-rs/oneil_output/src/error.rs
pixilcode: Yeah, once a file gets too big for me to navigate comfortably, I tend to split it up into smaller files. It speeds up development for me since I'm able to more effectively narrow down where I'm looking for something and I don't have to filter through as much content in a file.

Basically, it helps me to find things when code organization uses the file system rather than just the linear structure of text.

It makes sense to reduce the amount of file splitting in a PR, though. It can be hard to identify what is actually new content. Would it be better to submit a separate PR splitting a file before working with the split file? Also, do you believe that splitting this file is unnecessary?

README.md
pixilcode: That's a good point. I can make that clear.

src-rs/oneil_cli/src/lib.rs
pixilcode: Yeah, that would make sense. I'll add that in.

pixilcode:
> A lot of churn with the diagnostic change, which might have been better as a separate PR.

I agree. I started breaking up changes into smaller PRs with the three most recent PRs, but I didn't do it with [Fallback support](https://github.com/careweather/oneil/pull/36) (submitted about a month ago).

Most of your feedback was actually on code submitted for [Fallback support](https://github.com/careweather/oneil/pull/36), so I'll make the changes in that PR then rebase this and future PRs off of that. The only commits that are supposed to be for this PR are the last three commits. Alas, [Github stacked PRs](https://github.github.com/gh-stack/) aren't publicly available yet...

src-rs/oneil_output/src/error.rs
TimWhiting: This file is a bunch of object boilerplate. I would keep it together. I would split large files when they have implementation details concerning different objects that make it natural to split. If the objects are small and the functions are simple I would keep them together.

patrickwalton:
Doesn't look like there's any user side decisions I need to confirm here, so you can merge when ready.

PR #42, opened 2026-04-30
Add JSON serialization structure for the cache

pixilcode:
_NOTE: This PR is dependent on [Python import tracking](https://github.com/careweather/oneil/pull/41) and should not be reviewed until that PR is merged_

This PR adds data structures for serializing and deserializing the cache.


patrickwalton:
With @TimWhiting here, I'm not going to be reviewing code. I'll review at a user level, so in light of that can you add sample structure for the cache for me to take a look at or a spec?

pixilcode:
@patrickwalton Here is an example of what a cache would look like:

```json
{
  "imports": {
    "functions.py": {
      "name": "functions",
      "dependencies": ["math.py"],
      "hash": "139fa22dde725554",
      "functions_used": [
        "SX1276_sensitivity"
      ]
    }
  },
  "parameters": {
    "FSK_S": [
      {
        "function": "SX1276_sensitivity",
        "inputs": [
          {
            "value": 32.0,
            "unit": {
              "dimensions": {
                "bit": 1.0,
                "s": -1.0
              },
              "magnitude": 1000.0,
              "is_db": false,
              "display_unit": "kbps"
            }
          },
          {
            "value": 19999.999999999996,
            "unit": {
              "dimensions": {
                "s": -1.0
              },
              "magnitude": 1.0,
              "is_db": false,
              "display_unit": "rad/s"
            }
          },
          3.0
        ],
        "output": {
          "value": 2.462601799098393e-14,
          "unit": {
            "dimensions": {
              "kg": 1.0,
              "m": 2.0,
              "s": -3.0
            },
            "magnitude": 1.0,
            "is_db": false,
            "display_unit": "W"
          }
        }
      }
    ]
  },
  "tests": {
    "2": [
      {
        "function": "SX1276_sensitivity",
        "inputs": [
          {
            "value": 32.0,
            "unit": {
              "dimensions": {
                "bit": 1.0,
                "s": -1.0
              },
              "magnitude": 1000.0,
              "is_db": false,
              "display_unit": "kbps"
            }
          },
          {
            "value": 19999.999999999996,
            "unit": {
              "dimensions": {
                "s": -1.0
              },
              "magnitude": 1.0,
              "is_db": false,
              "display_unit": "rad/s"
            }
          },
          3.0
        ],
        "output": {
          "value": 2.462601799098393e-14,
          "unit": {
            "dimensions": {
              "kg": 1.0,
              "m": 2.0,
              "s": -3.0
            },
            "magnitude": 1.0,
            "is_db": false,
            "display_unit": "W"
          }
        }
      }
    ]
  }
}
```

src-rs/oneil_py_call_cache/README.md
```diff
@@ -0,0 +1,3 @@
+# oneil_cache
+
+TODO
```
TimWhiting: This will be finished prior to merge right.

src-rs/oneil_python/src/load.rs
```diff
@@ -83,6 +106,38 @@ fn insert_oneil_module_into_python(py: Python<'_>) -> PyResult<()> {
     Ok(())
 }
 
+// this adds the module directory to the sys.path so that the module can import
+// other modules in the module directory
+fn add_module_directory_to_sys_path(py: Python<'_>, module_directory: &Path) -> PyResult<()> {
```
TimWhiting, line 111: Can you rebase off of the latest so that we don't see already merged import tracking?

src-rs/oneil_py_call_cache/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "oneil_py_call_cache"
```
TimWhiting, line 2: Why a separate crate?

src-rs/oneil_py_call_cache/src/value.rs
```diff
@@ -0,0 +1,406 @@
...
+/// Converts an output interval into cache [`Interval`] form.
+const fn output_interval_to_cache(iv: oneil_output::Interval) -> Interval {
+    Interval {
+        min: iv.min(),
+        max: iv.max(),
+    }
+}
+
+/// Returns the JSON object key used for a [`Dimension`] exponent map.
+const fn dimension_to_cache_key(dim: Dimension) -> &'static str {
```
TimWhiting: Is there a reason for JSON to be human readable? I'm assuming no one will be actually looking at the caches? I would prefer to maintain less code. If we want a human-readable view of the cache, I'm assuming the extension could provide it. Also, the extension makes a lot of structures serde anyways, so we are going to end up with some conflicts likely.

pixilcode:
Hmm, that failing fuzz test doesn't have to do with the changes I made... It looks like interval arithmetic fails on a calculation of `sin` that I think is very specific. I'll see what I can do about it, but it shouldn't keep us from merging this branch. I think it's a bug that's been hidden in there for a while.

src-rs/oneil_py_call_cache/src/file.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /// Writes this cache as pretty-printed JSON to `path`, creating or truncating the file.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`WriteCacheError`] if the file cannot be created or JSON serialization fails.
+    ///
+    /// ```
+    /// use oneil_py_call_cache::FileCache;
```
TimWhiting: Can we remove the large doc comments here? We just need simple docs here.

src-rs/oneil_py_call_cache/src/identifiers.rs
```diff
@@ -0,0 +1,242 @@
...
+
+use oneil_shared::{
+    paths::PythonPath,
+    symbols::{ParameterName, PyFunctionName, TestIndex},
+};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// A path to a Python module.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct CachedPythonPath(PythonPath);
```
TimWhiting: Why all these newtypes / wrapper types? Paths are not cached. They might be part of something that is cached, but I don't know any difference in behavior between a cached path and non-cached path. It seems to me that the function call should uphold any invariants on this path.

TimWhiting: Same in the rest of this file.

src-rs/oneil_py_call_cache/src/value.rs
```diff
@@ -0,0 +1,406 @@
...
+use oneil_output::{
+    Dimension, DimensionMap, DisplayUnit, MeasuredNumber, Number, Unit as OutputUnit, Value,
+};
+use serde::de::{self, Deserializer};
+use serde::ser::Serializer;
+use serde::{Deserialize, Serialize};
+
+/// Wrapper around [`Value`] for JSON serialization and deserialization.
+#[derive(Debug, Clone, PartialEq)]
+pub struct CacheValue(pub Value);
```
TimWhiting: I also have to have deserialization for the LSP. I don't see what the issue is with just including serde on everything we need it on.

TimWhiting (changes requested):
I think this PR could be simplified quite a bit. Right now it introduces a lot of new code / maintainance burden that we don't need.

src-rs/oneil_python/src/load.rs
pixilcode: It has already been rebased. It only shows the two new commits for me. Maybe Github was showing you an old version of the PR for some reason?

src-rs/oneil_py_call_cache/src/value.rs
pixilcode: @patrickwalton specifically requested that the cache be human-readable. The features that he was originally looking for are listed [here](https://github.com/careweather/oneil/blob/main/README.md#function-caching) (written by Patrick, not myself). However, maybe things have changed since then. @patrickwalton, do you have feedback on this?

> Also, the extension makes a lot of structures serde anyways, so we are going to end up with some conflicts likely.

I don't think I understand what this means. Are you talking about conflicts with different versions of serde?

src-rs/oneil_py_call_cache/README.md
pixilcode: Oh, yes, thanks for catching that, it is definitely not complete :laughing:

src-rs/oneil_py_call_cache/Cargo.toml
pixilcode: It's a seperate concern. There wasn't a clear place to put this since it's a mechanism seperate from other features. I could put it in `oneil_runtime`, but then it would get confused with the other caches that the runtime uses. This cache is distinct from those other caches since it is written to the disk and persists between runs.

src-rs/oneil_py_call_cache/src/value.rs
pixilcode: Different features may need to serialize values differently. This is separate so that it doesn't lock other implementations into a specific representation.

src-rs/oneil_py_call_cache/src/identifiers.rs
pixilcode: The wrappers are so that we can implement `Serialize`/`Deserialize` on the values. Due to Rust's orphan rule, you need a newtype to implement those traits. It isn't intended to be passed around, it's main purpose is to serialize/deserialize the cache.

src-rs/oneil_py_call_cache/src/value.rs
patrickwalton: @TimWhiting typically a cache is meant for local usage only, but we're using caches a little differently here: to not require all users to have all simulation toolchains installed on their computer.

Remember I mentioned the status quo is to set requirements after someone runs one study? That's what we're trying to build an alternative too. For a very complex system, it's likely there are a lot of simulation suites and requiring everyone to install every simulation suite to run the model is burdensome -- people will probably still just leave the system engineering to someone else.

But if we can cache those simulation results, then other users run and modify the model as long as they don't impact inputs to the sim. When they do, a fallback is used and they get a warning that they should either install the simulation toolchain or have someone else re-run with their new inputs.

Because the cache is supporting collaboration here instead of just performance optimization, I think it's important that it's human readable.

src-rs/oneil_py_call_cache/src/file.rs
pixilcode: Yep, will do!

src-rs/oneil_python/src/load.rs
TimWhiting: I think I might have initially started the review this morning before the rebase.

src-rs/oneil_py_call_cache/Cargo.toml
TimWhiting: Okay, since it persists to disk, are we keeping track of an oneil version as well, so that we can verify / validate it was written with a version of oneil that the future reader understands.

src-rs/oneil_py_call_cache/src/value.rs
TimWhiting: Sounds fine to me

src-rs/oneil_py_call_cache/src/identifiers.rs
TimWhiting: Let's add serde support on the original types, we can wrap the ones that need to be human readable here.

src-rs/oneil_py_call_cache/Cargo.toml
pixilcode: I can definitely add that field!

src-rs/oneil_py_call_cache/src/value.rs
pixilcode: @TimWhiting, what did you mean by this? Are you talking about the version of the serde library?

> Also, the extension makes a lot of structures serde anyways, so we are going to end up with some conflicts likely.

TimWhiting: I mean my additions to the vscode extension with rendering requires many things to be serialized anyways. I guess we won't have too many conflicts, the changes will be mostly identical / easy to resolve.

src-rs/oneil_output/src/unit.rs
```diff
@@ -399,3 +473,103 @@ impl fmt::Display for DisplayUnit {
...
+    ///
+    /// This is to avoid re-parsing the display unit string.
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let name = String::deserialize(deserializer)?;
+        Ok(Self::Unit {
+            name,
+            exponent: 1.0,
```
TimWhiting: Is this lossy then, if the exponent is lost / defaulted?

src-rs/oneil_output/src/number_kinds/measured_number.rs
```diff
@@ -484,3 +486,88 @@ impl ops::Div<MeasuredNumber> for Number {
         }
     }
 }
+
+/// A measured number in a human-readable format.
+#[derive(Serialize, Deserialize)]
+struct HumanReadableMeasuredNumber {
```
TimWhiting, line 492: Must clearer intent here, thanks

TimWhiting (approved):
One small thing about unit exponents.

src-rs/oneil_output/src/unit.rs
pixilcode: Yes, though this is only for display. Unit information that is used for calculation is not lost.

I debated whether to encode the full display unit, but that would probably create more noise in the cache than would be worth it. After a function is called, the value's unit is often immediately reassigned by the parameter. And, even if the unit is displayed, the exponent won't show because the exponent is 1. So the display unit will display the same as if it were structured.

Do you think we should try and make it lossless, or should we keep it as is?

TimWhiting: It just seems wrong to me, to display a bad unit in the JSON if we want it to be human-readable.

pixilcode: The unit displayed in the JSON is displayed correctly, it's only when it's converted back from JSON that it might be displayed oddly. Serialization uses `to_string()`, which converts the display unit from a structured format to a string.

Also, note that it only _might_ be displayed oddly upon deserialization. In most cases, the display will still read correctly.

pixilcode: Maybe I could add another `DisplayUnit` variant called `Deserialized` or `Opaque` or something like that to indicate that it is an unparsed unit string?

TimWhiting: I think we should test a non-singleton exponent, and make sure we don't get weird round-trip issues. I can see how we might not -> it get's moved to the name and then stays in the name, but I'm a bit worried that name no longer is what it originally is supposed to be.

pixilcode: I added a test to demonstrate the behavior [here](https://github.com/careweather/oneil/pull/42/changes/279f4c84e409933884253d02803f2559ee816019#diff-a1e58568573e7555271b11f834118033a467e6f3b255b1d95ab6c5c62c57f293R576-R602). Does that make sense?

pixilcode: I've added a `RenderedUnit` variant to make it more clear what is happening with deserialization.

PR #43, opened 2026-04-30
Add caching for python function calls

pixilcode:
_NOTE: This PR is dependent on [the JSON serialization structure](https://github.com/careweather/oneil/pull/42) and should not be reviewed until that PR is merged._

This PR implements caching for python function calls in Oneil. Caches are stored in `__oncache__` in the directory from which `oneil` is run.

### DIFF
--- a/src-rs/oneil_eval/src/context.rs
+++ b/src-rs/oneil_eval/src/context.rs
@@ -49,11 +49,12 @@ pub trait ExternalEvaluationContext {
     /// Returns an error if there was an error evaluating the imported function.
     #[cfg(feature = "python")]
     fn evaluate_imported_function(
-        &self,
+        &mut self,
         python_path: &PythonPath,
         identifier: &PyFunctionName,
         function_call_span: Span,
         args: Vec<(output::Value, Span)>,
+        callsite_info: &CallsiteInfo,
     ) -> Option<Result<output::Value, Box<EvalError>>>;
 
     /// Returns a unit by name if it is defined in the builtin context.
@@ -225,6 +226,20 @@ enum SlotPeek {
     TakenForEval(Box<ir::Parameter>),
 }
 
+/// Information about the callsite of the function being evaluated.
+///
+/// This is used for caching Python function calls.
+#[derive(Debug, Default, Clone)]
+pub enum CallsiteInfo {
+    /// A parameter in a model.
+    Parameter(EvalInstanceKey, ParameterName),
+    /// A test in a model.
+    Test(EvalInstanceKey, TestIndex),
+    /// Other callsite (e.g. an expression in the REPL).
+    #[default]
+    Other,
+}
+
 /// Evaluation context that tracks per-instance memo state for lazy parameter evaluation.
 ///
 /// The context owns the parameter memo table for every instance the evaluator may touch:
@@ -245,6 +260,9 @@ pub struct EvalContext<'external, E: ExternalEvaluationContext> {
     eval_scope: Vec<EvalInstanceKey>,
     external_context: &'external mut E,
 
+    /// Context for the callsite of any function evaluated in the context.
+    callsite_context: Vec<CallsiteInfo>,
+
     /// Warnings for the parameter or test expression currently being evaluated.
     expression_eval_warnings: Vec<output::EvalWarning>,
 }
@@ -264,6 +282,7 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
             eval_scope: Vec::new(),
             external_context,
             expression_eval_warnings: Vec::new(),
+            callsite_context: Vec::new(),
         }
     }
 
@@ -292,6 +311,7 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
             models,
             eval_scope: Vec::new(),
             external_context,
+            callsite_context: Vec::new(),
             expression_eval_warnings: Vec::new(),
         }
     }
@@ -366,6 +386,7 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
             models,
             eval_scope: Vec::new(),
             external_context,
+            callsite_context: Vec::new(),
             expression_eval_warnings: Vec::new(),
         }
     }
@@ -617,7 +638,11 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
                     self.push_active_model(ak.clone());
                 }
 
-                let eval_result = eval_parameter::eval_parameter(&param, self);
+                let eval_result =
+                    eval_parameter::eval_parameter(parameter_name.clone(), &param, self);
+
+                // pop the callsite context the parameter evaluation
+                let _callsite_info = self.pop_callsite_context();
 
                 if let Some(ref ak) = anchor_key {
                     self.pop_active_model(ak);
@@ -698,7 +723,7 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
 
     /// Evaluates an imported function with the given arguments.
     pub fn evaluate_imported_function(
-        &self,
+        &mut self,
         python_path: &PythonPath,
         name: &PyFunctionName,
         function_call_span: Span,
@@ -706,8 +731,15 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
     ) -> Result<output::Value, Box<EvalError>> {
         #[cfg(feature = "python")]
         {
+            let callsite_info = self.callsite_context.last().cloned().unwrap_or_default();
             self.external_context
-                .evaluate_imported_function(python_path, name, function_call_span, args)
+                .evaluate_imported_function(
+                    python_path,
+                    name,
+                    function_call_span,
+                    args,
+                    &callsite_info,
+                )
                 .expect("imported function should be defined (checked during resolution)")
         }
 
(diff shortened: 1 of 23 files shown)


src-rs/oneil_eval/src/context.rs
```diff
@@ -617,7 +638,11 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
                     self.push_active_model(ak.clone());
                 }
 
-                let eval_result = eval_parameter::eval_parameter(&param, self);
+                let eval_result =
+                    eval_parameter::eval_parameter(parameter_name.clone(), &param, self);
```
TimWhiting, line 642: I noticed a lot of these cloning and addition of extra parameters. But the parameter itself is also a parameter. Does the param not have parameter_name?

src-rs/oneil_eval/src/context.rs
```diff
@@ -735,8 +767,31 @@ impl<'external, E: ExternalEvaluationContext> EvalContext<'external, E> {
...
+    pub fn begin_parameter_evaluation(&mut self, parameter_name: ParameterName) {
         self.expression_eval_warnings.clear();
+
+        let current_key = self
+            .eval_scope
+            .last()
+            .expect("current model should be set when beginning a parameter evaluation");
+
+        self.callsite_context
+            .push(CallsiteInfo::Parameter(current_key.clone(), parameter_name));
```
TimWhiting, line 779: How deep are needing the context? It seems like an additional parameter to parameter evaluation would be simpler, fewer invariants needed to be maintained?

src-rs/oneil_python/src/load.rs
```diff
@@ -85,8 +87,19 @@ pub fn load_python_import(
 
     // return the functions
     match functions {
-        Ok((module_docs, functions, imports)) => {
-            Ok(PythonModule::new(module_docs, functions, imports))
+        Ok((module_docs, functions, mut imports)) => {
+            // add the `requirements.txt` path to the imports if it exists
+            let maybe_requirements_txt_path = maybe_requirements_txt_path(&module_directory);
+            imports.extend(maybe_requirements_txt_path);
```
TimWhiting, line 93: It seems reasonable for now to have the requirements.txt be compared by textual equality, but this assumes that the packages installed by requirements.txt are the same versions, which will not be the case if it doesn't pin the versions. We should document this, and possibly run the install / hash the resolved versions of the dependencies if possible.

src-rs/oneil_python/src/source_hash.rs
```diff
@@ -0,0 +1,41 @@
...
+use crate::LoadPythonImportError;
+
+pub fn calculate_source_hash(source_paths: Vec<&Path>) -> Result<u64, LoadPythonImportError> {
+    // sort the source paths to ensure consistent hashing
+    let mut source_paths = source_paths;
+    source_paths.sort_unstable(); // `Path` implements `Ord` correctly, so `unstable` is okay
+
+    // hash the source files
+
+    let mut builder = Xxh3Default::new();
```
TimWhiting, line 15: This is not a cryptographically secure hash, so it can cause collisions, albeit rarely. We should at least add an implementation note, even if 99.9% of users won't need to worry about it.

src-rs/oneil_runtime/src/runtime/util.rs
```diff
@@ -25,6 +32,10 @@ impl Runtime {
             composed_graph: None,
             #[cfg(feature = "python")]
             python_import_cache: PythonImportCache::new(),
+            #[cfg(feature = "python")]
+            python_call_cache: PythonCallCache::new(cache_dir.clone()),
+            #[cfg(feature = "python")]
+            python_call_replacement_cache: PythonCallCache::new(cache_dir),
```
TimWhiting, line 38: Worth a note on why you have two caches? It's not clear to me why we need two. I assume that we either need to clear the full cache when invalidating the old one, or we evaluate / add one parameter at a time.

src-rs/oneil_runtime/src/cache.rs
```diff
@@ -193,94 +188,418 @@ impl EvalCache {
...
+            self.load(model_path).ok()?;
+
+            let entry = self.entries.get(model_path)?;
+            entry.parameters.get(parameter_name).map(Vec::as_slice)
+        }
+
+        pub fn get_test_entry(
+            &mut self,
+            model_path: &ModelPath,
+            test_index: TestIndex,
```
TimWhiting, line 381: Any reason why you don't give the test index / parameter entry in these functions and return an option of the result instead of option of list of results? It seems we should encapsulate that logic here, rather than at the call site. Otherwise, why not just use a map directly instead of going through this wrapper?

src-rs/oneil_runtime/src/cache.rs
```diff
@@ -193,94 +188,418 @@ impl EvalCache {
...
+        ///
+        /// Returns [`ReadCacheError`] if the cache file cannot be read.
+        fn load(&mut self, model_path: &ModelPath) -> Result<(), ReadCacheError> {
+            if self.entries.contains_key(model_path) {
+                return Ok(());
+            }
+
+            let cache_relative_path = get_cache_relative_path(model_path);
+            let cache_path = self.cache_dir.join(cache_relative_path);
+            let cache = FileCache::read_from_path(cache_path)?;
```
TimWhiting, line 460: Where do we validate that we are reading a compatible cache - written by an oneil version that our read supports?

src-rs/oneil_eval/src/context.rs
```diff
@@ -225,6 +226,20 @@ enum SlotPeek {
...
 
+/// Information about the callsite of the function being evaluated.
+///
+/// This is used for caching Python function calls.
+#[derive(Debug, Default, Clone)]
+pub enum CallsiteInfo {
+    /// A parameter in a model.
+    Parameter(EvalInstanceKey, ParameterName),
+    /// A test in a model.
+    Test(EvalInstanceKey, TestIndex),
```
TimWhiting, line 237: EvalInstanceKey is good, but we should also include the root model. Two different designs for the same model could have the same instance keys for submodels, but with different values. At least unless we include the design name as the root of the instancekey, which we don't do currently I think.

TimWhiting (changes requested):

src-rs/oneil_python/src/source_hash.rs
pixilcode: That's fair, a cryptographic hash would have a lower probability of collisions.

The probability is so low though (`2.92e-9` or `0.000000292%` for XXH3, calculated from their [collision testing](https://github.com/Cyan4973/xxHash/wiki/Collision-ratio-comparison#testing-64-bit-hashes-on-large-inputs-)) and it isn't that much greater than a cryptographic hash (`5.22e-10` or `0.0000000522%` for sha256, also calculated from their study) that I don't think that we will need to worry about it at all. I would guess that our users will never run into it.

src-rs/oneil_eval/src/context.rs
pixilcode: Good catch, it does. It makes sense not to duplicate the name. I'll fix that.

src-rs/oneil_python/src/source_hash.rs
TimWhiting: Okay, good to know. Sha256 in their table only uses part of the sha256 hash, so that isn't really cryptographically secure hash either. The fact that they found collisions means that collisions do happen, whereas the full sha256 hash is theoretically much harder to have a collision and there has never been found a sha256 collision in practice.

src-rs/oneil_eval/src/context.rs
TimWhiting: Just reread my comment, and I apologize for any confusion, there is a lot of parameters going on here.

src-rs/oneil_eval/src/context.rs
pixilcode: That makes sense, I'll pass the callsite context around as a local parameter instead.

src-rs/oneil_runtime/src/cache.rs
pixilcode: That's a good question. We currently don't check for a compatible cache. Should we have a "minimum acceptable Oneil version" and check that the version is greater than or equal to that? Then, anytime we update caching, we bump the minimum acceptable version to the current Oneil version? Or is there a better way to do it?

TimWhiting: I think that is a fine, that way if someone writes a cache with a new version of oneil with cache format changed and someone else tries to use it on an old version of oneil it reports an error, asks the user to upgrade or delete the cache.

src-rs/oneil_eval/src/context.rs
pixilcode: That sounds good. And it would make sense to keep the root model in the cache as well so that we can know when to invalidate function call entries.

src-rs/oneil_runtime/src/runtime/util.rs
pixilcode: Yeah, it's for the purpose of cache invalidation. The replacement cache is used to track which cache items are "fresh".

I'm thinking about moving that logic into the cache itself instead of having two `PythonCallCache`s. I think it would make cache coordination and invalidation easier. I'll reworking the cache.

src-rs/oneil_python/src/load.rs
pixilcode: Yeah, textual equality definitely doesn't imply that the underlying packages are the same, which means that the underlying package functionality could change from environment to environment.

It would make sense to hash the resolved versions of the dependencies, and I don't think it would be too hard to add, but this would essentially lock the cache into an exact set of dependencies.

I _think_ the idea is that you can use any of the defined set of dependencies and it is assumed that the result would be the same regardless. If that is the case, I agree that we should make sure to document that expectation for requirements.txt very clearly.

So I guess it just depends on how precisely we want to represent the "behavior" that is being cached.

pixilcode:
I needed to rework the cache so that cache entries were stored based on the python module that had the functions rather than the oneil model that functions are called from. See #49 for the replacement PR.

PR #45, opened 2026-05-01
Design documentation and more comprehensive documentation for submodel and references with grandchildren aliases.

TimWhiting:
This also adjusts the old keywords for submodel and reference in favor of the new, and adds an examples directory so that we can have samples to run to verify the correct output in the documentation.


pixilcode (approved):
The documentation looks good to me! Feel free to merge when ready.

PR #49, opened 2026-05-19
Python function caching v2.0

pixilcode:
My previous attempt at implementing caching (PR #43) resulted in complex code and bugs that were hard to track down. This is because of the way that the cache was structured. Cached calls were stored based on the model where they were called.

I started over with a new cache structure where cached calls are stored based on the python module that the function comes from. This allows for easier reuse of cached calls and better cache invalidation.

An example of the new cache structure is

```json
{
  "oneil_version": "0.16.0",
  "module_path": "script.py",
  "hash": "6dc70893e899a454",
  "dependencies": ["other.py"],
  "function_calls": {
    "addOne": [
      {
        "root_models": ["test_.on"],
        "inputs": [20.0],
        "output": 21.0
      }
    ]
  }
}

```

_I think I covered all of the feedback given in #43, but feel free to double check me on that_


pixilcode:
I will be implementing cache overwrite warnings and flags for turning caching on and off. The commits that I've made so far are ready for review, though.

pixilcode:
Okay, I've added prompting behavior and stale cache warnings. This should be everything for the PR.

_EDIT: I'm making a couple of small changes, but it still is mostly ready for review. Whenever you start reviewing, let me know and I'll start pushing my changes as actual commits rather than fixups_

src-rs/oneil_py_call_cache/src/function_call.rs
```diff
@@ -1,15 +1,21 @@
...
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct FunctionCall {
-    /// Name of the Python function.
-    pub function: PyFunctionName,
+    /// The root models that use this function
+    /// when evaluating themselves or their submodels.
+    ///
+    /// When this is empty, this function call should
+    /// be deleted from the cache.
+    pub root_models: BTreeSet<ModelPath>,
```
TimWhiting, line 18: Hmm, cache deletion seems like a tricky business, and it will be hard to get right, especially in the LSP setting where they open different files at different times. I would prefer to use a timestamping sort of method, and ask the user if they would like to remove an entry from the cache if it hasn't been used in awhile, or have a gc command that checks timestamps. Figuring out "when" a timestamp is stale is another tricky business - if it is the only element in the cache for that python file, it could be just a parameter that doesn't change much. We should probably only evict if the cache grows unwieldy (above a certain size, which could be configurable).

src-rs/oneil_py_call_cache/src/file.rs
```diff
@@ -61,8 +70,228 @@ impl FileCache {
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct Version {
+    pub major: u32,
+    pub minor: u32,
+    pub patch: u32,
+}
+
+impl FromStr for Version {
```
TimWhiting: I'm sure there is some semantic versioning package that we could use. If we ever use `-beta` or other semantic versioning details, we might regret building our own, not to mention it probably saves us a couple hundred lines of code.

src-rs/oneil_py_call_cache/src/file.rs
```diff
@@ -61,8 +70,228 @@ impl FileCache {
...
+                if !is_valid_version(&version) {
+                    return Err(de::Error::custom(format!(
+                        "unsupported oneil_version: {version}"
+                    )));
+                }
+                version
+            }
+            Some(_) => {
+                return Err(de::Error::custom(
+                    "oneil_version must be the first field in a cache file",
```
TimWhiting: JSON really doesn't care about ordering, which means JSON writers don't necessarily need to follow this convention. Worst case is we read a whole cache to determine that it needs to be thrown away. The simulation / calling python time will probably dominate anyways.

pixilcode: Yeah, this was something that I pondered for a while. The reason I chose to require `oneil_version` as the first field is that it ensures that any other parsing error is actually a case of invalid JSON, rather than just a new/outdated format.

For example, say we add a field `foo` to the cache format and we're trying to parse JSON in the old format (which is missing the `foo` field). If we try to parse the whole thing, _then_ check the version, parsing will fail with "cannot find field foo" rather than "using out-of-date version". Technically, both are valid errors, but I think "using out-of-date version" describes the problem more accurately than "cannot find field foo".

src-rs/oneil_py_call_cache/src/file.rs
pixilcode: That's fair, I could use the `semver` crate.

src-rs/oneil_py_call_cache/src/function_call.rs
pixilcode: I could see a timestamp-based strategy working. It's definitely a lot simpler of a solution than tracking the root models that reference a function call. To figure out when a timestamp is stale, we could refresh the timestamp every time the cached value is read, although that would then cause a lot of churn in the cache, which could be a problem if the cache is version-controlled.

I think that the root model idea should work, though, since evaluation should be deterministic (modulo python calls themselves). So if a call doesn't happen when a given root model is evaluated, then that call is likely gone from the model. Granted, this is not the case if the user is commenting out code for experimentation or something like that. Also, `foo() ? bar()` breaks this assumption if `foo()` only fails _sometimes_.

Also, I don't think we have to worry about the LSP, since it never writes to the cache, it only reads from it. Any writing that happens with the cache is done when the user runs a program.

Cache invalidation is definitely a problem I've been trying to figure out for a while, so I'm down to hear more thoughts on this.

src-rs/oneil_py_call_cache/src/file.rs
TimWhiting: This works, but technically new fields can be made optional instead to guard against this, and then a post validation pass ensures that the struct is the correct version (but old fields are locked in then). Alternatively we can use an enum to version, which handles the version tag for us, and additionally allows us to remove fields. E.g. a FileCacheV1 versus FileCacheV2 enum variant with an interface that allows for some compatibility between the two (providing default fields, etc), or version aware fallback or recovery, or just error if it isn't the right variant. The question then is whether we should depend on non-guaranteed JSON ordering or add an enum indirection, and extract the inner value for later consumption so we don't have to consider all variants throughout the code. Either way is more code to maintain, but the enum is more flexible, and serde can be derived automatically. What do you think?

src-rs/oneil_py_call_cache/src/function_call.rs
TimWhiting: Edited: After reading through the rest of the caching logic, the approach seems good currently.

However, we definitely need a way for the user to allow cache writes for the LSP, whether that is LSP configuration or some kind of interaction during evaluation.

src-rs/oneil_runtime/src/cache.rs
```diff
@@ -238,43 +292,987 @@ impl PythonImportCache {
...
+    /// Always read from the cache.
+    Always,
+    /// Never read from the cache.
+    Never,
+    /// Ask before reading from the cache.
+    Prompt(CachePrompterRef),
+}
+
+impl fmt::Debug for CacheReadPolicy {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
TimWhiting, line 804: Is there not a deriver for simple formatting like this?

src-rs/oneil_runtime/src/cache.rs
```diff
@@ -238,43 +292,987 @@ impl PythonImportCache {
...
+        }
+    }
+
+    /// Clear all entries that are no longer referenced by any root models.
+    fn clear_stale_entries(&mut self) -> IndexSet<PythonPath> {
+        for cache in self.entries.values_mut() {
+            // clear all cached function calls that are no longer
+            // referenced by any root models
+            for function_calls in cache.function_calls.values_mut() {
+                function_calls.retain(|call| !call.root_models.is_empty());
```
TimWhiting: retain on negation is hard to read switch to filter?

src-rs/oneil_runtime/src/cache.rs
```diff
@@ -238,43 +292,987 @@ impl PythonImportCache {
...
+            }
+
+            // clear all functions that have no cached function calls
+            cache
+                .function_calls
+                .retain(|_function_name, function_calls| !function_calls.is_empty());
```
TimWhiting: ditto

src-rs/oneil_runtime/src/cache.rs
```diff
@@ -238,43 +292,987 @@ impl PythonImportCache {
...
+            CacheWritePolicy::Always | CacheWritePolicy::Prompt(_) => (),
+        }
+
+        let mut errors = Vec::new();
+        let mut stale_caches = Vec::new();
+        for (python_path, cache) in &self.entries {
+            if self
+                .overwrite_outdated_caches
+                .get(python_path)
+                .is_none_or(|overwrite| *overwrite)
```
TimWhiting, line 621: So no explicit permission is permission? Also, this looks like it writes all entries, I would assume we only write updated/changed entries?

src-rs/oneil_py_call_cache/src/function_call.rs
pixilcode: Sounds good, I can make that happen.

src-rs/oneil_runtime/src/cache.rs
pixilcode: Unfortunately, as far as I can find, because `CachePrompterRef` can't derive Debug (since it's `Arc<dyn ...>`), it can't be derived for `CacheReadPolicy`. It would be nice if there were a `#[debug_ignore]` attribute or something like that.

src-rs/oneil_runtime/src/cache.rs
pixilcode: Yeah, I agree that it's harder to read than a `filter` function with no negation. I looked for a `filter` that behaved similarly to `retain` (one that doesn't require a new allocation), but I couldn't find one.

I could alternatively do

```rs
*function_calls = function_calls.iter().filter(|call| call.root_models.is_empty()).cloned();
```

but that would be allocating a new `Vec` and cloning all the function calls.

```rs
*function_calls = std::mem::take(function_calls).into_iter().filter(|call| call.root_models.is_empty());
```

is also an option. That would only allocate two `Vec`s.

I personally prefer `retain` over the other two options, since it feels readable enough, even though it isn't the ideal (it would be nice if `Vec` just had an in-place `filter`). I could also use `call.root_models.len() > 0` as the predicate if that would help readability.

TimWhiting: Open to the change in predicate I guess or just leaving it as is

src-rs/oneil_runtime/src/cache.rs
pixilcode: That's a good point. Since we're tracking which caches should be overwritten, this can be simplified to an iteration over `self.overwrite_outdated_caches`.

src-rs/oneil_runtime/src/cache.rs
pixilcode: Okay, I'll update the predicate.

src-rs/oneil_runtime/src/cache.rs
pixilcode: Actually, re-reading the code, the reason that I did it this way is that this also saves _new_ caches. If there wasn't a cache before, it needs to be saved. `self.overwrite_outdated_caches` is only populated with permissions for _overwriting_ caches. So if `self.overwrite_outdated_caches.get(python_path)` is `None`, that either means that the cache is new or the cache is unchanged. Either way, it can be written to disk without needing permission. So I believe "no explicit permission is permission" and "write all entries" is the correct behavior.

src-rs/oneil_py_call_cache/src/file.rs
pixilcode: I've converted the file cache to use an enum.

src-rs/oneil_py_call_cache/src/function_call.rs
pixilcode: Cache read/write policies can now be passed as arguments to the LSP. The extension also has options to configure this.

PR #52, opened 2026-06-09
Implement renaming in the LSP

pixilcode:
This implements renaming in the LSP for parameters and import aliases.

Note that this does _not_ cover renaming in design files. However, other LSP functionalities don't work in design files either because there is no IR associated with those files (at least that I could find). So I will implement LSP functionalities for design files in a seperate PR.


src-rs/oneil_lsp/src/rename.rs
```diff
@@ -0,0 +1,557 @@
...
+        }
+        ir::Limits::Discrete { values, .. } => {
+            for value in values {
+                collect_expr(model_path.clone(), value, mode, occurrences);
+            }
+        }
+    }
+}
+
+fn collect_expr(
```
TimWhiting: Do we not have a walk_expr or something that can visit / iterate over an expression/parameter definition which we could then pass a simple closure to to get the information we need? We should see if there is potential for deduplication / code reuse here.

src-rs/oneil_lsp/src/rename.rs
```diff
@@ -0,0 +1,557 @@
...
+            }
+        }
+    }
+}
+
+fn push_occurrence(occurrences: &mut Vec<RenameOccurrence>, model_path: ModelPath, span: Span) {
+    occurrences.push(RenameOccurrence { model_path, span });
+}
+
+fn dedupe_occurrences(mut occurrences: Vec<RenameOccurrence>) -> Vec<RenameOccurrence> {
```
TimWhiting: Are you aware why we have duplicated occurrences? Can we prevent it at another level (avoiding visiting a compilation unit more than once)?

src-rs/oneil_lsp/src/rename.rs
```diff
@@ -0,0 +1,557 @@
...
+        | SymbolAtPosition::PythonImport { .. }
+        | SymbolAtPosition::PythonFunctionReference { .. } => return None,
+    };
+
+    Some(PrepareRenameResponse::RangeWithPlaceholder { range, placeholder })
+}
+
+/// Returns the range to highlight during prepare-rename.
+#[must_use]
+fn prepare_rename_range(symbol: &SymbolAtPosition) -> Range {
```
TimWhiting: Inline this small method, only used one place.

src-rs/oneil_lsp/src/rename.rs
```diff
@@ -0,0 +1,557 @@
...
+    span_to_range(&symbol.span())
+}
+
+fn load_model<'runtime>(
+    runtime: &'runtime mut Runtime,
```
TimWhiting: Ditto, we can probably inline such a simple function

src-rs/oneil_lsp/src/rename.rs
```diff
@@ -0,0 +1,557 @@
...
+            o.model_path.clone(),
+            o.span.start().offset,
+            o.span.end().offset,
+        )
+    });
+    occurrences.dedup();
+    occurrences
+}
+
+fn sort_edits_descending(edits: &mut [TextEdit]) {
```
TimWhiting: Is this needed?

TimWhiting (dismissed):
A few items of feedback. I'd also like to see some attempt to do this for design files. Just try it out and see if it works. The ModelTemplateReference can be obtained for either model or design files as far as I can see.

src-rs/oneil_lsp/src/rename.rs
pixilcode: We do! I'll replace that here.

src-rs/oneil_lsp/src/rename.rs
pixilcode: No, reviewing it, I don't think there are duplicated occurrences. This is an artifact of doing the first pass with AI :sweat_smile:

src-rs/oneil_lsp/src/rename.rs
pixilcode: No, this is also an artifact of doing the first pass with AI.

pixilcode:
Fixed all the feedback!

> Just try it out and see if it works. The ModelTemplateReference can be obtained for either model or design files as far as I can see.

I did try this, but the `Model` IR is completely empty for a design.

pixilcode:
Looking into it further, the design surface is not captured in the design file itself, so nothing in the design surface can be directly renamed (or otherwise targeted by the LSP). I'm working on figuring out the best way to capture the design surface for rename/hover/etc., but I think it would be best in a seperate PR.

src-rs/oneil_lsp/src/lib.rs
```diff
@@ -410,9 +477,133 @@ impl LanguageServer for Backend {
 
         result
     }
+
+    async fn references(&self, params: ReferenceParams) -> Result<Option<Vec<Location>>> {
+        let position = params.text_document_position.position;
```
TimWhiting, line 484: Some repetition in this file, do these different params types implement a trait or something such that we can include the symbol at position / mutex lock into a higher order function that we can just pass a lambda for the specific functionality based on the position and any extra info from the parameters?

src-rs/oneil_cli/src/print_debug_ir.rs
```diff
@@ -201,6 +230,206 @@ enum SectionTag {
...
+        export_count += design.parameter_overrides().count();
+        export_count += design.scoped_parameter_overrides().count();
+        export_count += design.parameter_additions().count();
+    }
+    if config.sections.show_tests() {
+        export_count += design.test_additions().count();
+    }
+
+    if export_count > 0 {
+        section_list.push(("Design export", export_count, SectionTag::DesignExport));
```
TimWhiting, line 269: Is this count important? It seems to me that other information is more important. Not sure what this debug ir is used for though.

src-rs/oneil_lsp/src/occurrences.rs
```diff
@@ -0,0 +1,603 @@
...
+
+/// A single source occurrence.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Occurrence {
+    pub model_path: ModelPath,
+    pub span: Span,
+}
+
+/// Resolves a reference name to the model path it imports from a loaded model.
+pub fn resolve_reference_model_path(
```
TimWhiting: Is this the proper place for such helpers, it seems to me that this should be in the same module as ModelTemplateReference?

src-rs/oneil_lsp/src/occurrences.rs
```diff
@@ -0,0 +1,603 @@
...
+        ir::ParameterValue::Piecewise(exprs, _) => {
+            for piecewise in exprs {
+                collect_expr(model, target_model, piecewise.expr(), mode, occurrences);
+                collect_expr(model, target_model, piecewise.if_expr(), mode, occurrences);
+            }
+        }
+    }
+}
+
+fn collect_limits(
```
TimWhiting, line 411: Again we are iterating over models. Is there something preventing us from using a foreach / map / fold over the model?

src-rs/oneil_lsp/src/occurrences.rs
```diff
@@ -0,0 +1,603 @@
...
+                .design_export
+                .as_ref()?
+                .target_model()?;
+
+            (path == param_model_path).then(|| model.path().clone())
+        })
+        .collect()
+}
+
+fn push_occurrence(occurrences: &mut Vec<Occurrence>, model_path: ModelPath, span: Span) {
```
TimWhiting: This seems like something that should just be inlined.

src-rs/oneil_cli/src/print_debug_ir.rs
pixilcode: It just avoids printing out any design information if there isn't anything to print. This tool is mainly developer facing, I use it when debugging other problems.

src-rs/oneil_lsp/src/occurrences.rs
pixilcode: I could see us implementing that abstraction, that would be a good way to ensure that any changes to how we map/fold over a model only need to happen in one place. I think that the change would make sense in a seperate PR though, this PR has gotten pretty big already. I can add a task to the roadmap if you'd like!

TimWhiting: If you have time left in the week, and have finished the migration task, please do this now instead of deferring it to the roadmap. At least give it a try, if it turns out to be too large of a task we can defer it, but I think it would clean up all this new code, and make things easier to review.

pixilcode: I probably won't have it done this week since I'll be working on the migration all week. If it works as is, I think we should merge it now so that people can use the functionality, since people will likely be starting to use the new Oneil this week. Then, I can make this my next task after the migration, since I agree that this refactoring _would_ be an improvement to the code.

src-rs/oneil_lsp/src/occurrences.rs
```diff
@@ -0,0 +1,590 @@
...
+            && let Some(span) = alias_import.alias_span.as_ref()
+        {
+            let model_path = model.path().clone();
+            let span = span.clone();
+            occurrences.push(Occurrence { model_path, span });
+        }
+    }
+}
+
+fn get_designs_referencing_model(
```
TimWhiting, line 570: This function should move to runtime.

src-rs/oneil_lsp/src/occurrences.rs
TimWhiting: Please address the new comment instead. Then feel free to merge, I'll make sure to add this to your testing task.

PR #60, opened 2026-06-30
chore: Fix incorrect behavior found during migration

pixilcode:


pixilcode:
That's weird, git has stopped running githooks. I'll get that worked out, and I'll fix the broken tests.

pixilcode:
Tests are fixed, code is ready for review.

TimWhiting (approved):
Are there any other places where we serialized a float? I'm guessing there are and they need to be migrated to this serializer as well.

PR #65, opened 2026-07-14
chore: Improve tests

pixilcode:
This commit improves Oneil testing by:
- adding LSP tests
- adding evaluation tests
- adding analysis tests
- adding snapshot tests for tree and independents output

It also does some refactoring to improve the reusability of code and readability of tests.


src-rs/oneil_analysis/src/test_assertions.rs
```diff
@@ -0,0 +1,113 @@
...
+    assert!(
+        graph.validation_errors.is_empty(),
+        "expected no validation errors, got {:?}",
+        graph.validation_errors
+    );
+}
+
+/// Returns the dependency names of direct children, in tree order.
+#[must_use]
+pub fn child_dependency_names(tree: &Tree<DependencyTreeValue>) -> Vec<&DependencyName> {
```
TimWhiting: Should this function be in the impl on Tree?  It doesn't seem to be an assertion, just a helper?

src-rs/oneil_analysis/src/test_assertions.rs
```diff
@@ -0,0 +1,113 @@
...
+        .collect()
+}
+
+/// Asserts an independents map entry equals the expected `(name, value)` pairs.
+///
+/// # Panics
+///
+/// Panics if the maps differ.
+#[track_caller]
+pub fn assert_independent_params(
```
TimWhiting: What is this "independent" terminology? Independent parameter?

src-rs/oneil_snapshot_tests/src/snapshots/oneil_snapshot_tests__test__overlay_scoped_reference_override.snap
```diff
@@ -1,12 +1,11 @@
...
 ---
 Model: fixtures/design_overlay/scoped_ref_override/budget.on
 Tests: 0/0
 Parameters:
-  k = MeasuredNumber(MeasuredNumber { normalized_value: NormalizedNumber(Scalar(5.0000000000000004e-8)), unit: Unit { dimension_map: DimensionMap({Mass: 1.0, Distance: -1.0}), magnitude: 0.001, is_db: false, display_unit: Divide(Unit { name: "kg", exponent: 1.0 }, Unit { name: "km", exponent: 1.0 }) } })
-  m0 = MeasuredNumber(MeasuredNumber { normalized_value: NormalizedNumber(Scalar(15000.0)), unit: Unit { dimension_map: DimensionMap({Mass: 1.0}), magnitude: 1.0, is_db: false, display_unit: Unit { name: "kg", exponent: 1.0 } } })
-  m_req = MeasuredNumber(MeasuredNumber { normalized_value: NormalizedNumber(Scalar(20000.0)), unit: Unit { dimension_map: DimensionMap({Mass: 1.0}), magnitude: 1.0, is_db: false, display_unit: Unit { name: "kg", exponent: 1.0 } } })
-  m_margin = MeasuredNumber(MeasuredNumber { normalized_value: NormalizedNumber(Scalar(-5000.0)), unit: Unit { dimension_map: DimensionMap({Mass: 1.0}), magnitude: 1.0, is_db: false, display_unit: Unit { name: "kg", exponent: 1.0 } } })
+  k = 5e-5 :kg/km
```
TimWhiting, line 8: This one's value seemed to change? Or is it just that the magnitude of the unit hasn't been applied to the value yet? When do we do that?

src-rs/oneil_analysis/src/test_assertions.rs
```diff
@@ -0,0 +1,113 @@
...
+        .map(|child| &child.value().dependency_name)
+        .collect()
+}
+
+/// Returns kinds of all validation errors, in order.
+#[must_use]
+pub fn validation_error_kinds(
+    errors: &[InstanceValidationError],
+) -> Vec<&InstanceValidationErrorKind> {
+    errors.iter().map(InstanceValidationError::kind).collect()
```
TimWhiting: This can be inlined to usages.

src-rs/oneil_analysis/src/test_context.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    fn lookup_parameter_value(
+        &self,
+        instance_key: &EvalInstanceKey,
+        parameter_name: &ParameterName,
+    ) -> Option<Result<Parameter, GetValueError>> {
+        self.parameters
+            .get(&(instance_key.clone(), parameter_name.clone()))
+            .map(|result| match result {
```
TimWhiting: Is there not a way to just clone the inner value? It seems really annoying to have this unwrapping / rewrapping in an error value?

src-rs/oneil_eval/src/test_assertions.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/// Asserts that two floats are close (`is_close`).
+///
+/// # Panics
+///
+/// Panics if the values are not close.
+#[track_caller]
+pub fn assert_is_close(expected: f64, actual: f64) {
+    assert_is_close_case("", expected, actual);
```
TimWhiting: All of this delegation / repetition in this file could be reduced somehow I'm sure.

src-rs/oneil_snapshot_tests/src/util.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    if let Some(equation) = equation_str(display_info, file_cache) {
+        let pad = " ".repeat(name.len());
+        writeln!(out, "{indent}{rest_prefix}{pad} = {equation}")
+            .expect("String write is infallible");
+    }
+}
+
+fn tree_prefixes(depth: usize, is_first: bool) -> (&'static str, &'static str) {
+    if depth == 0 {
```
TimWhiting: I recall seeing tree building stuff elsewhere. Can we consolidate / reuse instead of duplicating?

src-rs/oneil_analysis/src/independents.rs
```diff
@@ -63,3 +63,245 @@ fn is_empty_dependencies(deps: &DependencySet) -> bool {
     //       considered to be constant
     deps.parameter_dependencies.is_empty() && deps.external_dependencies.is_empty()
 }
+
+#[cfg(test)]
+mod tests {
```
TimWhiting: This whole file is pretty hard for me to follow. I could probably figure it out given enough time. But maintaining it might be a struggle. Maybe make these snapshot tests? Try to only have a few files (1-3) for the fixtures, and just a lot of assertions about various parameters in the fixture.

This is probably a good thing to have in the code quality doc as well. We should reduce the fixture count as much as we can, and have multiple parameters in the fixtures. We shouldn't overdo it, but for something targeted like this independents, it probably makes sense to group it together. (Or reuse other fixtures from existing fixture patterns).

src-rs/oneil_eval/src/eval_expr.rs
```diff
@@ -510,3 +510,625 @@ fn eval_literal(value: &ir::Literal) -> Value {
...
+
+    mod binary_ops {
+        use super::*;
+
+        #[test]
+        fn eval_numeric_ops_table() {
+            let cases = [
+                (ir::BinaryOp::Add, 2.0, 3.0, 5.0),
+                (ir::BinaryOp::Sub, 5.0, 2.0, 3.0),
+                (ir::BinaryOp::EscapedSub, 5.0, 2.0, 3.0),
```
TimWhiting, line 585: Remind me is escaped sub supposed to have the same semantics as sub, or is it different - and if different we should probably test the difference. Same for escaped div.

src-rs/oneil_eval/src/eval_expr.rs
```diff
@@ -510,3 +510,625 @@ fn eval_literal(value: &ir::Literal) -> Value {
...
+        }
+    }
+
+    mod unit_cast {
+        use super::*;
+
+        #[test]
+        fn eval_cast_number_to_meters() {
+            let unit = ir_composite_unit([UnitSpec::new(None, Some("m"), false, 1.0)]);
+            let expr = ir::Expr::unit_cast(span(), lit_number(5.0), unit);
```
TimWhiting: Interesting I didn't realize we supported casting units? What is the syntax? There was someplace I needed this and couldn't get it to work.

src-rs/oneil_eval/src/test_fixtures.rs
```diff
@@ -0,0 +1,402 @@
...
+    };
+
+    units.fold(first, |acc, unit| {
+        ir::DisplayCompositeUnit::Multiply(Box::new(acc), Box::new(unit))
+    })
+}
+
+/// Builds an IR composite unit from unit specs (with a real display tree).
+#[must_use]
+pub fn ir_composite_unit(unit_list: impl IntoIterator<Item = UnitSpec>) -> ir::CompositeUnit {
```
TimWhiting: A few of these helper functions seem like they should not be in test fixtures, but in the place where the base type is defined.

src-rs/oneil_analysis/src/dependency.rs
```diff
@@ -547,3 +547,533 @@ fn resolve_external_instance_key(
 
     Some(current_key)
 }
+
+#[cfg(test)]
+mod tests {
+    use indexmap::IndexMap;
+    use oneil_ir as ir;
+    use oneil_output::Value;
+    use oneil_shared::{
```
TimWhiting: I would prefer we reuse existing fixtures for this. Hard to read.
All "correct behavior" should probably be able to use existing fixtures, and all "incorrect" should have a test (if there are few enough of them they can probably be here and not in a fixture).

TimWhiting (dismissed):
A few notes.

Primarily, let's try simplifying where we can. Although it is good to have more tests than fewer, it can get noisy quickly. Ideally most of our tests are about edge cases, and not just positive behavior. If we can reuse fixtures that would be great. Rust is pretty quick on tests, and fixtures are easier to comprehend then the test setup helpers / functions that we build here. A side effect is that they test the whole flow (lexing and parsing -> wherever the thing we are testing is). This is great for positive tests, but for negative / edge cases, we might not get to a later stage easily via fixtures. 

Make your changes on a separate commit. This is a lot of code, and I don't want to review these same changes again, just the diff.

src-rs/oneil_snapshot_tests/src/snapshots/oneil_snapshot_tests__test__overlay_scoped_reference_override.snap
pixilcode: A `MeasuredNumber` stores its value internally in a normalized form. This makes operations like `(5:km) + (10:m)` fast since no conversions need to take place before adding the normalized numbers. When it is displayed with the unit, however, it is converted from its normalized form to a number that is in terms of the unit. So `5.0000000000000004e-8` becomes `5.0000000000000004e-8 / 0.001 == 5e-5 :kg/km`.

src-rs/oneil_eval/src/eval_expr.rs
pixilcode: Yep! Unit casting syntax was added into the new Oneil. The syntax is `(<expr>:<unit>)` (ex. `(1:km)`).

src-rs/oneil_analysis/src/test_assertions.rs
pixilcode: Yeah, it refers to independent parameters, "independent" is the word that Patrick chose for it. It means that the parameter has no dependencies. It's used in the `independent` command.

For example, given `model.on`

```on
X: x = 1
Y: y = 2
Z: z = x + y
```

running `oneil independent model.on` would print out

```
x = 1
y = 2
```

src-rs/oneil_snapshot_tests/src/snapshots/oneil_snapshot_tests__test__overlay_scoped_reference_override.snap
pixilcode: The value hasn't changed, only how it's displayed. `MeasuredNumber` operates on values in a normalized form in order to make calculations involving the numbers faster. It only converts to the "correct" unit for display purposes.

EDIT: Sorry for the double post, I didn't see the first one and thought I had lost it.

src-rs/oneil_analysis/src/test_context.rs
pixilcode: I believe I've resolved this, though you're welcome to double check in case I misinterpreted what you meant. `lookup_parameter_value` now returns a `&Parameter` rather than a `Parameter`, so it avoids cloning until cloning is needed.

src-rs/oneil_eval/src/eval_expr.rs
TimWhiting: Hmm, I thought I tried this and it didn't work somewhere (if I recall correctly in a test expression for a constant unitless value). I'll have to give it another try.

src-rs/oneil_eval/src/eval_expr.rs
pixilcode: Escaped sub uses `min - min | max - max` instead of `min - max | max - min`. Same idea for escaped div. I can add some evaluation tests that use intervals.

src-rs/oneil_eval/src/eval_expr.rs
pixilcode: Let me know if it doesn't work and we can debug it.

pixilcode:
Okay, I think I've addressed all of your feedback. Let me know if there's anything I missed.

PR #66, opened 2026-07-14
chore(docs): Improve code quality docs

pixilcode:


docs/CODING_STANDARDS.md
```diff
@@ -158,6 +156,27 @@ fn bar(key: u32) -> u32 {
...
+```
+
+is better than
+
+```rs
+let reference_path = runtime
+    .load_model(model_path)
+    .references()
+    .get(reference_name)
+    .path();
```
TimWhiting: Even better, create a helper function for something like this / an impl on model. 

Then we have: 
`runtime.load_model(model_path).reference_path_of(reference_name)`
which is fine to be on one line.

In general when there is a chain of function calls, ask yourself if some segment of it can be made more generally useful throughout the codebase.

docs/CODING_STANDARDS.md
```diff
@@ -166,55 +185,129 @@ boilerplate, but they come at the price of a syntax that's harder to read and
...
+  alias_span: None,
+}
+
+Reference {
+  name: ReferenceName("foo"),
+  span: Span(7, 10),
+  // span with no name
+  alias_name: None,
+  alias_span: Some(Span(14, 15)),
+}
```
TimWhiting: We can reduce the example noise overhead. If you have an example of bad vs good, and a good explanation, don't include extra examples explaining the bad (just get to the point) - look throughout this file for the same issue. If you want and can find a blog post explaining the particular issue in more detail, that would be more useful than adding our own explanations.

docs/CODING_STANDARDS.md
```diff
@@ -256,13 +349,85 @@ match value {
...
+  let a = input.0;
+  let b = input.1;
+  assert_eq!(a + b, b + a);
+})
+```
+
+The unit test is deterministic and only runs once, whereas the property test
+is run on many semi-random inputs. Running on many inputs enables a property
+test to discover inputs that break the property, if any exist. This can give
+you confidence that a property is probably sound.
```
TimWhiting: probably sound is too programming languages specific terminology. Just say that `confidence that a property holds`. Confidence already implies no proof.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+Do not use magic numbers. Always show your work or your sources. Clarify your
+assumptions.
+
+### Subdivide Into Hierarchical Subsystems
+
+Subdivide models into logical hierarchical subsystems. You should typically
+align these subsystems with a specific hardware component if it stands by
+itself. If a functionality is filled collaboratively by multiple subsystems, it
+should be modeled in a top-level system model.
+
```
TimWhiting, line 16: top-level is probably not neccessary.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+Specify the design inputs and calculate the performance output, not the other
+way around. Independent parameters (those that are assigned a value instead of
+an equation) should generally be design parameters that the engineer has more
+direct control over.
+
+### One Source of Truth
+
+Do not duplicate parameters. There should be one source of truth for each
+physical property or relationship. If this is not possible for some reason, use
+comments to make clear that this is a duplicate parameter.
```
TimWhiting: It should always be possible, or we should make sure that we can make it possible. It is an organizational / modeling issue.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+### IDs
+
+Parameter IDs should be as simple as possible. Prefer short subscripts and don't
+use multiple subscripts (`v_wmx` instead of `v_wind_max`).
+
+IDs are used to produce typeset equations. The shorter the name the better. For
+example, battery voltage should use `V_b` instead of `V_batt`.
+
+In typesetting, imported submodels are given as a superscript. If the battery
```
TimWhiting: Currently in the renderer I think the submodel is a `[model]` or something. This should reflect that for now, how we typeset them in latex can come later when we support parametric documents.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+Be very clear in the note that follows the parameter. Provide a description of
+how you derived the equation or obtained a value. Provide sources where
+relevant, either URLs or journal references. But do not repeat yourself. For
+example, if the parameter name is "Flux capacitor power consumption", don't say
+in the note "This is the power consumption of the flux capacitor"; instead say,
+"taken from the Doc's own Delorean handbook, page 13."
+
+Write notes in LaTeX. If you give a URL in a note, use `\href`. Escape special
+LaTeX characters like `%` and `&`.
```
TimWhiting: Take a look at the renderer, we use different citation format, we use markdown for headings / emphasis / bold / tables. The only place we use latex in notes should be equations.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+If multiple parameters would give the same URL as a source, consider including
+that source in the introductory note and referencing it in the parameter notes.
+For example, if this is an off-the-shelf electronic component, the introductory
+note would give the source for the datasheet and the parameter notes could just
+say something like, "given on page # of the datasheet."
```
TimWhiting: Citations can actually reference the page of the citation (see the renderer), it would be good to have this documented. I believe we use Pandoc citation style, with a few extra goodies.

See http://madoko.org/reference.html#sec-bib 4.12.2 Citation styles for a short explanation of some of the supported styles. We should document those here.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Don't Repeat Yourself
+
+For Oneil, name, ID, math, units, and sources/notes all have their own place:
+
+- Don't put units in the name, ID, or note.
+- Don't re-state the name in the note.
+- Don't re-state the math in the note, unless you derive it in more detail
```
TimWhiting, line 104: Here is a good place to put that you can interpolate equations and values in the text {{R:equation}}, {{R:value}}, the rendered view will have the latex rendered equation and values in those places.

An exception to repeating yourself is for summary tables (this is where the interpolation would most commonly go - it shouldn't be used in the note for the equation itself probably).

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+It's generally better to structure your submodels around actual hardware, at
+least the lowest-level models, because then you can have a model file that's
+tied to the specifications and properties of one component.
+
+For example, if you have a `solar.on` file which represents a solar power
+system, it could import a `SM500K12L.on`, which represents a specific solar cell
+component that can be purchased off the shelf. If a Oneil file refers
+specifically to an off-the-shelf component, it is preferable to name the file
+after the component model number.
```
TimWhiting: Alternatively a potentially better way to structure this is to also have a `solar_cell.on` that models the essential aspects of a solar cell, and for SM500K12L.on to be instead a design file, in particular if we might switch out something later.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Constants
+
+If a parameter is a fact that is generally true regardless of the component or
+design, include it in a `constants.on` file and import it. For example, the
+speed of light should go in `constants.on`.
```
TimWhiting: Make sure to mention that we should always import constants.on as `reference` not `submodel`.

docs/ONEIL_CODING_STANDARDS.md
```diff
@@ -0,0 +1,143 @@
...
+
+Use tests to model relationships between parameters. For example, let's say you
+are designing a smartphone. You specify the battery capacity, `C_b`, and the
+model calculates the corresponding battery volume, `V_b`. You could use a
+relational test to make sure the battery volume is not larger than the total
+smartphone volume, `V`:
+
+```oneil
+test : V_b < V
+```
```
TimWhiting, line 181: Note that every performance parameter should have at least one test referencing it (best practice).

docs/CODING_STANDARDS.md
```diff
@@ -256,13 +349,85 @@ match value {
...
+
+For snapshot testing in Rust, Oneil uses the `insta` crate. Snapshot tests are
+found in [`oneil_snapshot_tests`](../src-rs/oneil_snapshot_tests/).
+
+When writing a snapshot test, make the output easy to read. Don't use default
+`Debug` output. For example,
+
+```plaintext
+5 | 10 :kg
+```
```
TimWhiting, line 350: Let's audit our current snapshot tests and migrate them if they aren't this way. I know I created a few the other way - maybe some of them test things that are hard to see in plaintext, but we should update those that we can.

TimWhiting (changes requested):
Looks pretty good, let's make sure we keep things concise and to the point. Let's also add the Rust principle that we should not use getters and setters and should just use the structs directly as much as possible. That will clean up a lot of code.

docs/CODING_STANDARDS.md
pixilcode: Yeah, this was the case in the snapshot tests. I updated it in the PR where I improve the tests, specifically in [this commit](https://github.com/careweather/oneil/pull/65/changes/139af1788d8ff34bc6336fc6def29e93568f3e81).

docs/CODING_STANDARDS.md
pixilcode: Yeah, I can probably get rid of one of these examples, one probably makes the point. I think the bad examples are useful, though, since they give a point of comparison. If you can't identify what "bad code" looks like, you can't replace it with good code.

I do agree that having a blog post that explains the concept more in detail would be helpful. The only thing I've been able to find that does it justice is [The Life of a File](https://youtu.be/XpDsk374LDE), but that's more oriented towards Elm and it's a video, so it probably wouldn't work well in these docs. I've also found [this article](https://corrode.dev/blog/illegal-state/) that is about Rust, but it doesn't quite capture the same idea.

TimWhiting: The problem with including bad code is sometimes people look for recommended practices and they just copy the samples. We don't want them accidentally copying the wrong example. Similarly AI sometimes pays attention to the patterns that are most repeated. Imagine looking at the name of this file (CODING_STANDARDS.md) and looking for the most common patterns in the file. If most of the patterns are bad, that is bad input for both humans and AI particularly when the file name is CODING STANDARDS. Comparisons between good and bad are probably better suited to a linting or best practices docs, not a standards doc.

TimWhiting:
I would make sure that we have included all of Patrick's oneil advice, and probably ask for a review from him on the Oneil doc.

pixilcode:
Sounds good. The Oneil standards are drawn almost directly from [the ruleset](https://careweather.github.io/oneil/b-using-ai.html) that Patrick wrote for using Oneil with AI, and I can ask him for a review as well.

TimWhiting (approved):

TimWhiting:
Please accordingly update the ruleset as well then, fix typos and remove absolute (not general) paths from there while you are at it.

pixilcode:
Done! The linting failure is taken care of in the test improvements branch (Rust recently released a new version, which catches more lints). I can rebase if you'd like, though. Otherwise, let me know if there's anything else that you find!

PR #79, opened 2026-07-23
fix(validation): Validate limits when validating the parameter

### LINKED ISSUE #72: Missing quotes in limit strings fail silently.
Describe the bug
Unquoted identifiers in discrete parameter limits are treated as parameter references, but Oneil does not validate expressions inside parameter limits. Evaluation fails with hidden internal errors, prints no model output, and exits with status 0.

To Reproduce

Create repro.on:
$ Choice [foo, bar]: choice = 'foo'
Run:
oneil repro.on -P all
echo $?
Observe no useful diagnostic and exit status 0.
Run:
oneil repro.on -D -P all --dev-show-internal-errors
Observe hidden ParameterHasError diagnostics for foo and bar.
Observed behavior
oneil check repro.on reports no problem. Normal evaluation suppresses the internal errors and model output, then exits successfully with status 0.

Expected behavior
Oneil should report undefined parameters in limit expressions during validation, identify foo and bar at their source locations, and exit nonzero.

If discrete string limits require quotes, the diagnostic could suggest:

$ Choice ['foo', 'bar']: choice = 'foo'
Screenshots
Not applicable; terminal output can be included as text.

Environment:

OS: Fedora Linux 43, kernel 7.0.12-101.fc43.x86_64
Oneil version: v0.16.1
Relevant files: crsra.on, oneil_analysis/src/validation.rs, oneil_frontend/src/instance/graph.rs, oneil_cli/src/print_error.rs
Additional context
Parameter values and tests are traversed during variable classification and validation, but parameter limits are not. Evaluation subsequently produces ParameterHasError, which is marked internal and hidden by default. The CLI still detects the hidden error and suppresses model output, but does not return a nonzero exit status.

pixilcode:
Validates that variables used in limits are defined. Closes #72 .

### DIFF
--- a/src-rs/oneil_analysis/src/validation.rs
+++ b/src-rs/oneil_analysis/src/validation.rs
@@ -146,12 +146,24 @@ fn validate_instance(
         let provenance = parameter.design_provenance();
         let anchor = provenance
             .and_then(|prov| resolve_anchor_for_validation(instance, ancestors, &prov.anchor_path));
+        let param_scope = anchor.unwrap_or(instance);
         let design_info =
             provenance.map(|prov| (prov.design_path.clone(), prov.assignment_span.clone()));
         validate_value(
             parameter.value(),
             instance,
-            anchor.unwrap_or(instance),
+            param_scope,
+            host_path,
+            &location,
+            design_info.as_ref(),
+            pool,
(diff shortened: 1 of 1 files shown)


TimWhiting (approved):
LGTM

PR #81, opened 2026-07-27
chore(ci): Add Github Action for weekly quality review

pixilcode:
This should work in theory, but I haven't tested it at all. I also don't know if it needs to be on the `main` branch to work. Also, this doesn't follow the principle of least privilege since it gives `write` permissions for `contents`, `pull-requests`, and `issues` to everything, rather than just the parts that need it.


TimWhiting:
I think this is a great place to start.

PR #83, opened 2026-08-10
feat(cli): Add oneil test --format json and fix its exit code

TimWhiting:
This allows for simple CI checks for automated testing of Oneil models and determining regressions.


TimWhiting:
Addressed coding-standards review findings in a5af5883:

- Integration test now asserts the full serialized JSON schema via `serde_json::json!` equality (with `model_path` normalized to the basename so the contract isn't host-path dependent).
- `build_report_is_a_success_when_every_test_passes` asserts `report.models.len() == 1` before indexing.
- Text-format path in `handle_test_command` hoists `TestPrintConfig` and uses an explicit `skip_model_results` flag to flatten control flow.

TimWhiting:
@patrickwalton, I'll merge rust-rewrite into main after merging this in, and update the changelog / version tag / etc.

